PR #1, opened 2017-10-10
[Do not merge] Ideas how to use apq

elsid:


ideas/single_connection_usage_with_tuple_cursor.cpp
```diff
@@ -0,0 +1,12 @@
+boost::asio::io_service io_service;
+
+boost::asio::spawn(io_service, [&] (boost::asio::yield_context yield) {
+    auto connection_provider = apq::make_connection_provider(io_service);
+    apq::cursor<std::tuple<std::string>> cursor;
+    apq::binary_protocol::request(connection_provider, cursor, "SELECT pg_is_in_recovery()", yield);
+    std::tuple<std::string> data;
+    apq::binary_protocol::request(cursor, data, yield);
```
thed636: Here must be ```fetch``` of course

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+
+/**
+* Should we use defined type of pg_connect_handle, or some customized type and
+* template for connection?
+*   PGconn                       - pointer to native connection handle
+*   std::function<void(PGconn*)> - finalization function, e.g. destructor for single 
+*                                  connection or recycling operation for pool.
+*/
+using pg_connect_handle = std::shared_ptr<PGconn>; //, std::function<void(PGconn*)>>;
```
elsid, line 11: I think we definitely shouldn't use ```std::function``` for finalization. For both standalone and polled connection we know finalization functions. Let's use some customized type or two different types.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+
+/**
+* We can try to use standard error_code, but we will loose the error message
+* from database. But if we will use the custom error_code we are forced to
+* implement our own yield_context and async_result infrastructure. This perspective
+* is not attractive. We can provide some kind of yield_context support, but with
+* message loss in case of error_code bound to yield (i.e. yield[ec])
+*/
+using error_code = boost::system::error_code; // std:: / custom ?
```
elsid, line 21: Until we use ```boost::asio``` let's use ```boost::system_error```. For networking ts we'll back to this question.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+template <typename ConnectionProvider, typename CompletionToken>
+auto get_connection(ConnectionProvider&& p, CompletionToken&& token) {
+	using detail::get_connection_impl;
+	detail::async_result_init<CompletionToken, void (error_code, connection)> init(std::move(token));
+	get_connection_impl(std::forward<ConnectionProvider>(p), init.handler);
+	return init.result.get();
+}
+
+
+struct dsn; // or just an alias to std::string?
```
elsid, line 58: Alias is enough.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+
+class connection_pool {
+public:
+	template <typename Handler>
+	void get_connection(io_service& ios, Handler&& h);
+};
+
+// Returns ConnectionProvider for specified dsn
+auto make_connection_provider(io_service&, dsn);
```
elsid, line 68: Not only dsn, but config.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+auto make_connection_provider(io_service&, dsn); 
+
+// Return ConnectionProvider for specified connection pool
+auto make_connection_provider(io_service& ios, connection_pool& pool) {
+	return [](auto&& handler) {
+		pool.get_connection(ios, std::move(handler));
+	}
+}
+
+enum class parameters_transfer_mode {binary, text};
```
elsid, line 77: It is both the parameters and the result, may be just ```protocol_type```?

ideas/apq.h
thed636: Well, it doesn't matter which standard code we use std or boost, the main problem is when we use standard error_code we loose an error message form PostgreSQL. We can use custom error_code like this prototype in my repo https://github.com/thed636/error_code/blob/master/include/error_code/error_code.h. But we need to provide some mechanism to support our internal mail_error::error_code.

ideas/apq.h
thed636: There is a trade-off between std::function (or any other abstraction) and several types. It is about connection class. In case of the different types we need to use a template of the connection class. And this leads to templating of the transaction_connection class, commit and rollback functions. IMO this fact makes code slightly more complicated. But I can be wrong.

ideas/apq.h
thed636: I'm totally agree with you.

ideas/apq.h
thed636: Yes, good idea!

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+	template <typename Handler>
+	void get_connection(io_service& ios, Handler&& h);
+};
+
+// Returns ConnectionProvider for specified dsn
+auto make_connection_provider(io_service&, dsn); 
+
+// Return ConnectionProvider for specified connection pool
+auto make_connection_provider(io_service& ios, connection_pool& pool) {
+	return [](auto&& handler) {
```
thed636, line 72: ```return [&ios](auto&& handler) {``` of course.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+	template <typename T, typename CompletionToken>
+	static auto fetch(cursor<T>& c, T& t, CompletionToken token);
+
+	template <typename ConnectionProvider, typename CompletionToken>
+	static auto update(ConnectionProvider provider, query q, CompletionToken token);
+
+	template <typename ConnectionProvider, typename CompletionToken>
+	static auto execute(ConnectionProvider provider, query q, CompletionToken token);
+
```
thed636: I missed a ```prepare(...)``` function here.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+* message loss in case of error_code bound to yield (i.e. yield[ec])
+*/
+using error_code = boost::system::error_code; // std:: / custom ?
+
+/**
+* Connection - represents binding of hative handle an io_service (io_context) object
+* Should we allow to copy connection itself, or the object must be move-only?
+*/
+struct connection {
+	io_service& ios;
```
elsid, line 28: No tabs in this project, please.

ideas/single_connection_usage_with_timeouts.cpp
```diff
@@ -2,12 +2,13 @@ boost::asio::io_service io_service;
...
 boost::asio::spawn(io_service, [&] (boost::asio::yield_context yield) {
     auto connection = apq::make_connection_provider(io_service);
-    apq::cursor cursor;
+    apq::data_row row;
     connection_timeouts timeouts;
     timeouts.connect = 100ms;
     timeouts.request = 200ms;
-    apq::text_protocol::request(connection, cursor, timeouts, "SELECT pg_is_in_recovery()", yield);
-    std::cout << cursor.next(yield).at(0) << '\n';
+    if (apq::text_protocol::request(connection, apq::make_query("SELECT pg_is_in_recovery()", timeouts), row, yield)) {
```
elsid: Why timeouts are the part of query?

elsid: What is the result of request, why we should use it as condition? Besides, there is always one row result in this example.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+template <typename T>
+using cursor = std::unique_ptr<void(T&, void (error_code, size_type))>;
+
+class transaction_connection;
+
+/**
+* Represents raw data row in case of request.
+*/
+class data_row;
```
elsid, line 87: Why not just ```row```?

elsid:
Headers with ```.hpp``` extension, please.

ideas/apq.h
thed636: I'll fix it. Didn't cate about tabs for the prototype. :)

ideas/apq.h
thed636: I can't remember the actual reasons why not - let it be just ```row``` :)

thed636:
> Headers with .hpp extension, please.

I don't think what ```.hpp``` is a great idea due to naming convention suggestion from the [C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Rs-file-suffix). If You have no technical facts to force us to use ```.hpp``` instead of ```.h``` then I'd prefer to use ```.h``` extension according to the Guidelines.

ideas/single_connection_usage_with_timeouts.cpp
```diff
@@ -0,0 +1,15 @@
+boost::asio::io_service io_service;
+
+boost::asio::spawn(io_service, [&] (boost::asio::yield_context yield) {
+    auto connection = apq::make_connection_provider(io_service);
+    apq::cursor<apq::row> cursor;
+    connection_timeouts timeouts;
+    timeouts.connect = 100ms;
+    timeouts.request = 200ms;
+    apq::text_protocol::request(connection, apq::make_query("SELECT pg_is_in_recovery()", timeouts), cursor, yield);
```
thed636, line 9: About timeouts. In this case we allow to specify custom timeouts for the query. It is useful feature and IMO it is better to provide it rather than force a user to make his (or hers) own solution.

ideas/apq.h
dpotepalov: >  when we use standard error_code we loose an error message form PostgreSQL

why? can we not communicate that message through a custom error_category?

dpotepalov:
Okay, a bit late to the party, but still:
* do we really want to return `auto` from library interface functions? IMHO it impairs readability for library users
* why completion handlers are called CompletionToken? For me a token is something authentication-related and definetly not callable
* @thed636 - why is working with ConnectionProvider inside the library is better than working with it on the outside? Right now it seems that the first thing every request-related function does is get_connection from provider. Why not accept a connection right in `request(...)`, for example? All connection-related strategies can be accessed via separate functions, and the user may use any strategy that he/she wants to obtain a connection of any particular type (single, pooled, lazy etc.)

ideas/apq.h
thed636: > why? can we not communicate that message through a custom error_category?

Nope - error_category can not provide an ability for context depended information such as error messages from database which contains information about syntax error in query's text for example. So you can say "oh, you have a syntax error" with standard error_code, but you can not say "oh, you have a syntax error at column 45". "at column 45" is a context depended information, and there is no rooms for it in std or boos error_code. What's why we are using custom error_code in the meta access.

thed636:
> do we really want to return auto from library interface functions? IMHO it impairs readability for library users

It depends on. If the result type depends on arguments type then auto is better than use complicated and ugly helpers for types definition. If result type isn't depended form arguments type it is better to point concrete type, of course. E.g. 'request()' function return type depends on CompletionToken type - for callback and yield context it will be different types.

> why completion handlers are called CompletionToken? For me a token is something authentication-related and definetly not callable

E.g. boost::asio::yield_context or boost::asio::use_future are not handlers at all.

> why is working with ConnectionProvider inside the library is better...

In most cases user is interested in single operation, so in this case with connection provider code will be:

    request(pool, ..., handler);

and with naked connection:

    get_connection(pool, [handler](error_code ec , auto connection) {
        if(ec) {
            handler(ec);
        } else {
            request(connection, ..., handler);
        }
    });

In first case you will get and hold connection right before request execution, in second - you will get connection earlier and potentially hold connection for a longer time. Moreover, as you can see from the example, user will write more code in common cases. 

With first approach you still give an ability to hold and manage connection by user for special cases, but with common cases it is easer than second approach. Pay attention, that connection is a connection provider itself, so user can provide connection as an argument of the functions.

thed636:
> Okay, a bit late to the party

I'm glad to see you here :)

dpotepalov:
> E.g. boost::asio::yield_context or boost::asio::use_future are not handlers at all.

Ah, yes. I rarely used futures, so auto return type and CompletionToken looked weird to me). I get it now.
> connection provider itself, so user can provide connection as an argument of the functions.

Thanks for the verbose explanation). So the provider helps the client to reduce boilerplate whilst still allowing custom strategies. And it can be reasonably assumed that the compiler will inline get_connection and the call to handler in case of single connection, so there is no need for specializations with ConnectionProvider = connection. Nice).

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+
+template <parameters_transfer_mode prm>
+struct basic_protocol {
```
thed636, line 91: It looks like we do not need text protocol in case of binary protocol properly implemented. So we can simplify this part.

ideas/apq.h
```diff
@@ -0,0 +1,285 @@
...
+
+make_query( "SELECT name FROM my_table WHERE id="_sql << id << " AND sid=" << sid );
+
+make_query( "SELECT name FROM my_table WHERE id=:id  AND sid=:sid"_sql % (("id",id), ("sid",sid)) );
+
+
+BOOST_FUSION_DEFINE_STRUCT((), (my_query_params), 
+	(std::string, id)
+	(int, sid)
+);
```
thed636, line 264: We need some type definition mechanism to allow map C++ types to database types through the OIDs and type names. So we need some traits system with easy macro for custom types definition.  We can predefine all the traits for geninue types like int, string and so on. E.g. for composite types we need something like BOOST_FUSION_DEFINE_STRUCT but with type traits definition inside, let it be something like this:

    APQ_DEFINE_COMPOSITE(<namespace>, <schema>, <type-name>, <composite-fields>);
    APQ_DEFINE_CUSTOM_TYPE(<namespace>, <schema>, <type-name>, <type-definition>);
    APQ_DEFINE_ENUM(<namespace>, <schema>, <type-name>, <enum-items>);

Another problem is UTF-8 string representation. Do we need to provide some custom type of validated UTF-8 or just use std::string?

ideas/apq.h
```diff
@@ -0,0 +1,362 @@
...
+template <>
+inline std::size_t type_size(const std::string& v) noexcept {
+    return v.size();
+}
+
+/**
+* Oid acquisition for types must looks like (simplified):
+* auto oid = type_oid(v, [oid_map](const auto& v) {
+*         return oid_map.at(type_name(v));
+*     });
```
thed636, line 349: This case can be built with a compile-time check. It requires preparing something like boost::fusion::map with a key by type. So values of such "map" can be filled during connecting to database, or on error with invalidated oid.

ideas/apq.h
dpotepalov: um... I tend to disagree. Text protocol is vastly simpler and can be always used as fallback in case something goes wrong with the binary (old server, new server, bug in some rarely used type etc). libpq is flexible enough to allow to choose representation for each query parameter and for the result also. Shouldn't we provide at least as much flexibility?

thed636: The text protocol and binary protocol are differing only in a way how to represent parameters. In case of binary protocol, the parameters represented as is in a binary way and marked with OIDs. In text protocol, you need to serialize all values as strings manually (including escaping which is not trivial in some cases and so on) point their types in a query mandatory and at the end pass them as a binary representation of strings. Do we need such a protocol support?

thed636:
No more needed

PR #2, opened 2017-11-03
Add dummy test and build scripts

elsid:


README.md
```diff
@@ -1 +1,36 @@
...
+
+[![Build Status](https://travis-ci.org/elsid/apq.svg?branch=master)](https://travis-ci.org/elsid/apq)
+
+## Build
+
+### With custom environment
+
+Requirements:
+* CMake
+* C++ and C compilers
```
dpotepalov: also Boost?

tests/example.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Example
```
dpotepalov: come to think of it, do we really need Gherkin in this project? The library is pretty technical, tests written in C++ should describe it's behavior quite succinctly, and the target audience will be able to read them.

dpotepalov (commented):
looks good, but Gherkin examples need to be replaced with real tests or removed

tests/example.feature
thed636: It is not a matter of target audience - the main concept is to describe functionality with natural language, to combine documentation and tests tightly because in another way you need to interpret test and sometimes it can be not so easy to do it. The natural language description is more user-friendly than a bunch of tests with very weird names in most cases. So this is all about functional or acceptance tests. Of course, it is not necessary to use Gherkin for unit testing such us serialization mechanism or query config parsing functionality.

Anyway, in case we understand what this is overengineering we will remove it.

README.md
elsid: yes

tests/example.feature
elsid: We can use Gherkin (GUnit allows), if we find reason, but for now I will remove example. It still available [here](https://github.com/cpp-testing/GUnit/blob/master/example/GScenario.feature).

thed636 (approved):
We need to start from some point :)

PR #3, opened 2017-12-08
Type traits system

thed636:
Add type traits system primitives which can be used for writing query parameters and reading binary data.

### DIFF
--- a/include/apq/apq.h
+++ b/include/apq/apq.h
@@ -1,3 +1,7 @@
 #pragma once
 
 #include <apq/connection_pool.h>
+
+namespace libapq {
+
+} // namespace libapq
--- a/include/apq/query.h
+++ b/include/apq/query.h
@@ -2,7 +2,11 @@
 
 #include <boost/hana.hpp>
 
+namespace libapq {
+
 template <const char* text, class ... Args>
 auto make_query(Args&& ... args) {
     return boost::hana::make_tuple(std::forward<Args&&>(args) ...);
 }
+
+} // namespace libapq
--- a/include/apq/type_traits.h
+++ b/include/apq/type_traits.h
@@ -176,14 +176,14 @@ struct type_traits_helper {
 };
 
 /**
-* By default all non specialized types become with unspecified name, name
+* By default all non specialized types have unspecified name, name
 * oid and size.
 */
 template <typename T>
 struct type_traits : type_traits_helper<void> {};
 
 /**
-* Condition indicates if the specified typr is built-in for PG
+* Condition indicates if the specified type is built-in for PG
 */
 template <typename T>
 struct is_built_in : std::integral_constant<bool,
--- a/include/apq/type_traits.h
+++ b/include/apq/type_traits.h
@@ -1,0 +1,317 @@
+#pragma once
+
+#include <apq/detail/pg_type.h>
+
+#include <boost/hana/at_key.hpp>
+#include <boost/hana/insert.hpp>
+#include <boost/hana/string.hpp>
+#include <boost/hana/map.hpp>
+#include <boost/hana/pair.hpp>
+#include <boost/hana/type.hpp>
+#include <boost/uuid/uuid.hpp>
+#include <boost/optional.hpp>
+#include <boost/scoped_ptr.hpp>
+#include <boost/shared_ptr.hpp>
+#include <boost/weak_ptr.hpp>
+// Fusion adaptors support
+#include <boost/fusion/support/is_sequence.hpp>
+#include <boost/fusion/include/is_sequence.hpp>
+
+#include <memory>
+#include <string>
+#include <list>
+#include <vector>
+#include <set>
+#include <type_traits>
+
+namespace libapq {
+
+namespace hana = ::boost::hana;
+using namespace hana::literals;
+/**
+* PostgreSQL OID type - object identificator
+*/
+using oid_t = int32_t;
+
+/**
+* Type for non initialized OID
+*/
+using null_oid_t = std::integral_constant<oid_t, 0>;
+constexpr null_oid_t null_oid;
+
+/**
+* Indicates if type is nullable. By default - type is not
+* nullable.
+*/
+template <typename T>
+struct is_nullable : ::std::false_type {};
+
+/**
+* These next types are nullable
+*/
+template <typename T>
+struct is_nullable<::boost::optional<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::std::optional<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::boost::scoped_ptr<T>> : ::std::true_type {};
+template <typename T, typename Deleter>
+struct is_nullable<::std::unique_ptr<T, Deleter>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::boost::shared_ptr<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::std::shared_ptr<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::boost::weak_ptr<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::std::weak_ptr<T>> : ::std::true_type {};
+
+template <typename T>
+inline auto is_null(const T& v) noexcept ->
+    std::enable_if_t<is_nullable<std::decay_t<T>>::value, bool> {
+    return !v;
+}
+
+template <typename T>
+inline auto is_null(const ::boost::weak_ptr<T>& v) noexcept {
+    return is_null(v.lock());
+}
+
+template <typename T>
+inline auto is_null(const ::std::weak_ptr<T>& v) noexcept {
+    return is_null(v.lock());
+}
+
+template <typename T>
+constexpr auto is_null(const T&) noexcept ->
+    std::enable_if_t<!is_nullable<std::decay_t<T>>::value, std::false_type> {
+    return {};
+}
+
+/**
+* Indicates if type is an array
+*/
+template <typename T>
+struct is_array : ::std::false_type {};
+
+template <typename T, typename Alloc>
+struct is_array<::std::vector<T, Alloc>> : ::std::true_type {};
+template <typename T, typename Alloc>
+struct is_array<::std::list<T, Alloc>> : ::std::true_type {};
+
+/**
+* Helper indicates if type is string
+*/
+template <typename T>
+struct is_string : std::false_type {};
+
+template <typename C, typename T, typename A>
+struct is_string<std::basic_string<C, T, A>> : std::true_type {};
+
+/**
+* Indicates if type is adapted for introspection with Boost.Fusion
+*/
+template <typename T>
+using is_fusion_adapted = std::integral_constant<bool, 
+    ::boost::fusion::traits::is_sequence<T>::value
+>;
+
+/**
+* Indicates if type is adapted for introspection with Boost.Hana
+*/
+template <typename T>
+using is_hana_adapted = std::integral_constant<bool,
+    ::boost::hana::Sequence<T>::value ||
+    ::boost::hana::Struct<T>::value
+>;
+
+/**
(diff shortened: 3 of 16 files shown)


include/apq/query.h
```diff
@@ -2,7 +2,11 @@
...
 #include <boost/hana.hpp>
 
+namespace libapq {
+
 template <const char* text, class ... Args>
 auto make_query(Args&& ... args) {
     return boost::hana::make_tuple(std::forward<Args&&>(args) ...);
 }
+
+} // namespace libapq
```
elsid, line 12: You may just delete this file. It had been commited accidentally.

include/apq/apq.h
```diff
@@ -1,3 +1,7 @@
 #pragma once
 
 #include <apq/connection_pool.h>
+
+namespace libapq {
```
elsid, line 5: Why libapq?

tests/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Werror")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Werror -DBOOST_HANA_CONFIG_ENABLE_STRING_UDL=1")
```
elsid: Let's do not mix defines and flags in one string literal.

scripts/build_clang_debug.sh
```diff
@@ -7,7 +7,7 @@ BUILD_DIR=${BUILD_PREFIX}build_clang_debug
 mkdir -p ${BUILD_DIR}
 cd ${BUILD_DIR}
 cmake \
-    -DCMAKE_CXX_FLAGS='-std=c++17 -Wall -Wextra -pedantic -Werror' \
+    -DCMAKE_CXX_FLAGS='-std=c++1z -Wall -Wextra -pedantic -Werror' \
```
elsid: Why `1z`?

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+            Size\
+        >{};\
+    }
+
+LIBAPQ_PG_DEFINE_TYPE(bool, "bool", BOOLOID, bytes<1>)
+LIBAPQ_PG_DEFINE_TYPE(char, "char", CHAROID, bytes<1>)
+LIBAPQ_PG_DEFINE_TYPE(std::vector<char>, "bytea", BYTEAOID, dynamic_size)
+
+LIBAPQ_PG_DEFINE_TYPE(boost::uuids::uuid, "uuid", UUIDOID, bytes<16>)
+LIBAPQ_PG_DEFINE_TYPE(std::vector<boost::uuids::uuid>, "[]uuid", 2951, dynamic_size)
```
elsid: Why not `uuid[]`?

elsid (changes requested):

include/apq/apq.h
thed636: Name "apq" conflicts with the internal project name. Do you have any naming suggestions?

include/apq/query.h
thed636: Let it be a placeholder :)

include/apq/type_traits.h
thed636: Oops - my mistake. I'll fix it.

scripts/build_clang_debug.sh
thed636: Mac OS stock compiler has no such option like C++17. But I can revert this change.

tests/CMakeLists.txt
thed636: Ok

scripts/build_clang_debug.sh
elsid: Too bad for mac. Revert, please.

include/apq/apq.h
elsid: `yapq`?

scripts/build_clang_debug.sh
thed636: Well, if we want to be user-friendly we must support it - nobody needs a product to compile it only for docker image ;) I'll fix it a little bit later.

include/apq/apq.h
thed636: is it really better? :)

dpotepalov: how about pqaxx for namespaces, libpqaxx for the project? Includes a not very subtle "axe"-pun in it... But the real reason I'm suggesting this is because I've googled around for other pq-libraries, and they all seem to be named libpq*xx: libpqxx, libpqmxx, libpq++...

tests/CMakeLists.txt
```diff
@@ -25,6 +28,7 @@ link_directories("${source_dir}/libs/gherkin-cpp")
 
 set(SOURCES
     connection_pool.cpp
+    type_traits.cpp
```
dpotepalov, line 31: @thed636 I think you force-pushed the wrong commit as head of the branch, I don't see this file or anything actually related to traits in the PR

thed636: Oops, my mistake. I've pushed the branch one more time.

include/apq/apq.h
thed636: We do not provide adaptation of libpq. We provide an advanced convenient mechanism for asynchronous PostgreSQL queries. In this case, "apq" is the best name. But unfortunately, we have a project with this name. I don't like "lib" prefix, to be honest, and I thought about something like apq2, but such name will confuse users. I don't think what to name our project similar to the other projects is a good idea. In this case, a user can't easily distinguish this project from many other ones. It is better to have an original name.

How about "bpq - Better PostgreSQL Queries"? :) Or "acma_pq" with short namespace "acma" (sounds really good!)? Or "amq" - advanced mechanism [for] queries? Or even "pqnyan" - in this case, we can use pretty cat-mascot as a logo? =(^_^)=

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+* These next types are nullable
+*/
+template <typename T>
+struct is_nullable<::boost::optional<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::std::optional<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::boost::scoped_ptr<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::std::unique_ptr<T>> : ::std::true_type {};
```
dpotepalov: unique_ptr with custom deleter is also nullable

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+struct is_nullable<::std::shared_ptr<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::boost::weak_ptr<T>> : ::std::true_type {};
+template <typename T>
+struct is_nullable<::std::weak_ptr<T>> : ::std::true_type {};
+
+template <typename T>
+inline auto is_null(const T& v) noexcept ->
+    std::enable_if_t<is_nullable<std::decay_t<T>>::value, bool> {
+    return !v;
```
dpotepalov, line 72: will not work for weak_ptr, specialization which locks prior to checking is required

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+}
+
+/**
+* Indicates if type is an array
+*/
+template <typename T>
+struct is_array : ::std::false_type {};
+
+template <typename T>
+struct is_array<::std::vector<T>> : ::std::true_type {};
```
dpotepalov: with custom allocator - also an array

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+template <typename T>
+struct is_array<::std::list<T>> : ::std::true_type {};
+template <typename T>
+struct is_array<::std::set<T>> : ::std::true_type {};
```
dpotepalov: I understand what are you trying to accomplish, but this statement feels just plain wrong. With list it's sort of okay, but set is not suitable for accepting an array of values, since an array can contain duplicates, and set cannot. Therefore if the database returns an array and user supplies a set to us, we will essentially lie about the result.

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+struct is_composite : std::integral_constant<bool,
+    ::std::is_class<T>::value &&
+    !is_nullable<T>::value &&
+    !is_array<T>::value &&
+    !is_string<T>::value
+> {};
+
+
+/**
+* Type traits template forvard declaration. Type traits contains information
```
dpotepalov: forWard

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+* related to it's representation in the database. There are two different
+* kind of traits - built-in types with constant OIDs and custom types with
+* database depended OIDs. The functions below describe neccesary traits.
```
dpotepalov: dependeNT

include/apq/type_traits.h
```diff
@@ -0,0 +1,287 @@
...
+template <typename T>
+struct is_string : std::false_type {};
+
+template <typename C, typename T, typename A>
+struct is_string<std::basic_string<C, T, A>> : std::true_type {};
+/**
+* Indicates if type is a composite
+*/
+template <typename T>
+struct is_composite : std::integral_constant<bool,
```
dpotepalov: Is this trait really necessary? It seems to return true for a lot of types and probably lies about some of them. For example, any user-defined type that is not a composite is composite, judging by this trait.

include/apq/type_traits.h
thed636: **Done**

include/apq/type_traits.h
thed636: **Done**

include/apq/type_traits.h
thed636: **Done**

include/apq/type_traits.h
thed636: Yes, you right. I think we should allow to represent it like an array for serialization but with different type compatibility mechanism.

include/apq/type_traits.h
thed636: Yes, you partially right, I think the better predicate is to determine if user adapt his type via boost::hana (or maybe boost::fusion as legacy). But anyway, a user always can specialize this template for his (or hers) type.

include/apq/type_traits.h
thed636: **Done**

include/apq/type_traits.h
thed636: **Done**

I need a spellchecker for comments :) @elsid can we get some pre-commit grammar-nazi bot? :)

scripts/build_clang_debug.sh
thed636: **Done**

thed636 (commented):
Uploaded new version

include/apq/type_traits.h
thed636: **Done**

dpotepalov (approved):
I fixed some errors in comments, decided that would be faster than leaving messages in PR. Besides those, looks ok to me.

include/apq/type_traits.h
```diff
@@ -0,0 +1,317 @@
...
+
+template <typename T>
+using is_dynamic_size = std::is_same<typename type_traits<T>::size, dynamic_size>;
+
+/**
+* Function returns type name in Postgre SQL. For custom types user must
+* define this name.
+*/
+template <typename T>
+constexpr auto type_name(const T&) noexcept {
```
elsid, line 200: We need function that returns `boost::hana::string` to support compile time query text generation.

thed636: If you are talking about query for custom type oids then I think it is better to use this for every type in the oid map:

    typename type_traits<T>::name{};

elsid (approved):

PR #5, opened 2017-12-16
Query

elsid:
Partial implementation of libpq query constructor with binary serialization and query text build. Now supports only arithmetic types and strings, but review of the concept is required. Based on type_traits branch rebased onto master.

### DIFF
--- a/include/ozo/binary_serialization.h
+++ b/include/ozo/binary_serialization.h
@@ -1,0 +1,174 @@
+#pragma once
+
+#include <ozo/concept.h>
+#include <ozo/type_traits.h>
+
+#include <libpq-fe.h>
+
+#include <boost/hana/adapt_struct.hpp>
+#include <boost/hana/for_each.hpp>
+#include <boost/hana/members.hpp>
+#include <boost/hana/tuple.hpp>
+#include <boost/range/algorithm/for_each.hpp>
+
+#include <type_traits>
+
+#include <limits.h>
+
+namespace ozo {
+namespace detail {
+
+struct pg_array {
+    BOOST_HANA_DEFINE_STRUCT(pg_array,
+        (std::int32_t, dimensions_count),
+        (std::int32_t, dataoffset),
+        (::Oid, elemtype)
+    );
+};
+
+struct pg_array_dimension {
+    BOOST_HANA_DEFINE_STRUCT(pg_array_dimension,
+        (std::int32_t, size),
+        (std::int32_t, index),
+        (std::int32_t, begin)
+    );
+};
+
+template <class T, std::size_t ... N>
+constexpr T byte_order_swap(T value, std::index_sequence<N ...>) {
+    return (((value >> N * CHAR_BIT & std::numeric_limits<std::uint8_t>::max()) << (sizeof(T) - 1 - N) * CHAR_BIT) | ...);
+}
+
+template <class T, class U = std::make_unsigned_t<T>>
+constexpr Require<endian::native == endian::little, U> convert_to_big_endian(T value) {
+    return byte_order_swap<U>(value, std::make_index_sequence<sizeof(T)> {});
+}
+
+template <class T>
+constexpr Require<endian::native == endian::big, T> convert_to_big_endian(T value) {
+    return value;
+}
+
+template <class T, class OutIteratorT>
+constexpr Require<std::is_integral_v<T>, OutIteratorT> write(T value, OutIteratorT out) {
+    hana::for_each(
+        hana::to<hana::tuple_tag>(hana::make_range(hana::size_c<0>, hana::size_c<sizeof(T)>)),
+        [&] (auto i) { *out++ = value >> decltype(i)::value * CHAR_BIT & std::numeric_limits<std::uint8_t>::max(); }
+    );
+    return out;
+}
+
+template <class T>
+struct floating_point_integral {};
+
+template <>
+struct floating_point_integral<float> {
+    using type = std::uint32_t;
+};
+
+template <>
+struct floating_point_integral<double> {
+    using type = std::uint64_t;
+};
+
+template <class T>
+using floating_point_integral_t = typename floating_point_integral<T>::type;
+
+template <class T, class OutIteratorT>
+constexpr Require<std::is_floating_point_v<T>, OutIteratorT> write(T value, OutIteratorT out) {
+    union {
+        T input;
+        floating_point_integral_t<T> output;
+    } data {value};
+    return write(data.output, out);
+}
+
+template <class T>
+constexpr Require<std::is_arithmetic_v<T>, std::size_t> get_size(const T&) noexcept {
+    return sizeof(T);
+}
+
+template <class T>
+constexpr Require<!std::is_arithmetic_v<T>, std::size_t> get_size(const T& value) noexcept {
+    return std::size(value);
+}
+
+template <class OutIteratorT>
+auto make_writer(OutIteratorT& out) {
+    return [&] (auto v) { out = write(v, out); };
+}
+
+template <class OidMapT, class OutIteratorT>
+auto make_sender(const OidMapT& oid_map, OutIteratorT& out) {
+    return [&] (const auto& v) { out = send(v, oid_map, out); };
+}
+
+} // namespace detail
+
+template <typename T>
+constexpr auto SingleByteIntegral = std::is_integral_v<T> && sizeof(T) == 1;
+
+template <class T, class OidMapT, class OutIteratorT>
+constexpr Require<SingleByteIntegral<T>, OutIteratorT> send(T value, const OidMapT&, OutIteratorT out) {
+    return detail::write(value, out);
+}
+
+template <typename T>
+constexpr auto MultiByteIntegral = std::is_integral_v<T> && (sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8);
+
+template <class T, class OidMapT, class OutIteratorT>
+constexpr Require<MultiByteIntegral<T>, OutIteratorT> send(T value, const OidMapT&, OutIteratorT out) {
+    return detail::write(detail::convert_to_big_endian(value), out);
+}
+
+template <typename T>
+constexpr auto FloatingPoint = std::is_floating_point_v<T> && (sizeof(T) == 4 || sizeof(T) == 8);
+
+template <class T, class OidMapT, class OutIteratorT>
+constexpr Require<FloatingPoint<T>, OutIteratorT> send(T value, const OidMapT&, OutIteratorT out) {
+    return detail::write(value, out);
+}
+
+template <class OidMapT, class OutIteratorT, class CharT, class TraitsT, class AllocatorT>
+constexpr OutIteratorT send(const std::basic_string<CharT, TraitsT, AllocatorT>& value, const OidMapT&, OutIteratorT out) {
+    boost::for_each(value, detail::make_writer(out));
+    return out;
+}
+
+template <class OidMapT, class OutIteratorT, class T, class AllocatorT>
+constexpr Require<std::is_arithmetic_v<T>, OutIteratorT> send(const std::vector<T, AllocatorT>& value, const OidMapT& oid_map, OutIteratorT out) {
+    out = send(detail::pg_array {1, 0, ::Oid(type_oid<std::decay_t<T>>(oid_map))}, oid_map, out);
+    out = send(detail::pg_array_dimension {std::int32_t(std::size(value)), 0, 1}, oid_map, out);
+    boost::for_each(value,
+        [&] (const auto& v) {
+            out = send(std::int32_t(detail::get_size(v)), oid_map, out);
+            out = send(v, oid_map, out);
+        });
+    return out;
+}
+
+template <class OidMapT, class OutIteratorT, class AllocatorT>
+constexpr OutIteratorT send(const std::vector<char, AllocatorT>& value, const OidMapT&, OutIteratorT out) {
+    boost::for_each(value, detail::make_writer(out));
+    return out;
+}
+
+template <class OidMapT, class OutIteratorT>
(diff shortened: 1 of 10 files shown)


include/apq/query.h
```diff
@@ -1,12 +1,184 @@
...
-    return boost::hana::make_tuple(std::forward<Args&&>(args) ...);
+template <class T>
+using not_nullopt_t = std::enable_if_t<!std::is_same_v<T, std::nullopt_t>, T>;
+
+template <class T>
+using is_nullable_t = std::enable_if_t<is_nullable<std::decay_t<T>>::value, void>;
+
+template <class T>
+using is_not_nullable_t = std::enable_if_t<!is_nullable<std::decay_t<T>>::value, void>;
+
```
thed636: A minute of advertising. :)

With Require from connection branch this can be rewritten like:

    template <typename T>
    constexpr auto Nullable = is_nullable<std::decay_t<T>>::value;

...

    template <class T>
    static constexpr Require<!Nullable<T>> write_meta(const oid_map_type& oid_map, ...)

thed636 (commented):
IMO it is better to use query concept with default implementation instead of concrete class for better integration ability. I think the best practice is to define a functional set for the concept to give a user an ability to built own query class as it is needed. E.g. first use case is our own integration with current query implementation. In this case, we can integrate with current solution step by step.

include/ozo/query_builder.h
```diff
@@ -0,0 +1,193 @@
...
+    using query_element_param = query_element<std::reference_wrapper<const ValueT>, query_param_tag>;
+    using tail_tuple = hana::tuple<query_element_param>;
+    return make_query_builder(hana::concat(std::move(builder.elements), tail_tuple(query_element_param {value})));
+}
+
+namespace literals {
+
+template <class CharT, CharT ... c>
+constexpr auto operator "" _SQL() {
+    namespace hana = boost::hana;
```
thed636: We have such alias in **type_traits.h**, do we need a header with Hana shortcut defined and common headers included?

include/ozo/binary_serialization.h
```diff
@@ -0,0 +1,138 @@
...
+    std::for_each(value.begin(), value.end(),
+        [&] (const auto& v) {
+            send(detail::get_size(v), oid_map, out);
+            send(v, oid_map, out);
+        });
+}
+
+template <class OidMapT, class OutIteratorT, class AllocatorT>
+constexpr void send(const std::vector<char, AllocatorT>& value, const OidMapT&, OutIteratorT out) {
+    std::for_each(value.begin(), value.end(), [&] (auto v) { detail::write(v, out); });
```
thed636: Don't you like to use range-based **boost::for_each**?

include/ozo/binary_serialization.h
```diff
@@ -0,0 +1,138 @@
...
+}
+
+template <class OidMapT, class OutIteratorT, class T, class AllocatorT>
+constexpr void send(const std::vector<T, AllocatorT>& value, const OidMapT& oid_map, OutIteratorT out) {
+    send(std::int32_t(detail::get_dimension(value)), oid_map, out);
+    send(std::int32_t(0), oid_map, out);
+    send(type_oid<T>(oid_map), oid_map, out);
+    send(std::int32_t(std::size(value)), oid_map, out);
+    send(std::int32_t(1), oid_map, out);
```
thed636: Would you like to make a sequence like a tuple and use **hana::for_each** algorithm?

    const auto header = hana::make_tuple(
        std::int32_t(detail::get_dimension(value)),
        std::int32_t(0),
        type_oid<T>(oid_map),
        std::int32_t(std::size(value)),
        std::int32_t(1));

    hana::for_each(header, [&](auto x) { send(x, oid_map, out);});


Maybe it is better to define header structure via **BOOST_HANA_DEFINE_STRUCT** and use **hana::for_each** then. But it seems like the solution comes with a little bit more boilerplate code.

include/ozo/binary_serialization.h
```diff
@@ -0,0 +1,138 @@
...
+}
+
+template <class OidMapT, class OutIteratorT, class AllocatorT>
+constexpr void send(const std::vector<char, AllocatorT>& value, const OidMapT&, OutIteratorT out) {
+    std::for_each(value.begin(), value.end(), [&] (auto v) { detail::write(v, out); });
+}
+
+template <class OidMapT, class OutIteratorT>
+constexpr void send(const boost::uuids::uuid& value, const OidMapT&, OutIteratorT out) {
+    std::for_each(value.begin(), value.end(), [&] (auto v) { detail::write(v, out); });
```
thed636: Since **uuid** is a fixed size type you can try to use **hana::for_each** here.

include/ozo/impl/query.h
```diff
@@ -0,0 +1,186 @@
...
+    basic_query(const char* text)
+        : impl(make_impl(buffer_allocator_type(), text, oid_map_type())) {}
+
+    basic_query(const char* text, const not_nullopt_t<oid_map_type>& oid_map, const ArgsT& ... args)
+        : impl(make_impl(buffer_allocator_type(), text, oid_map, args ...)) {}
+
+    basic_query(const buffer_allocator_type& buffer_allocator, const char* text, const not_nullopt_t<oid_map_type>& oid_map, const ArgsT& ... args)
+        : impl(make_impl(buffer_allocator, text, oid_map, args ...)) {}
+
+    constexpr const char *text() const noexcept {
```
thed636: This comment is related to the query concept. I think it is better to expect ranges than raw pointers. In pointer case, user's query concept implementation must hold all this data inside itself. This provides additional restrictions for the user implementation, which are not convenient and can force the user to modify own query class instead of adaptation with free functions.

include/ozo/binary_serialization.h
```diff
@@ -0,0 +1,138 @@
...
+using floating_point_t = std::enable_if_t<std::is_floating_point_v<T> && (sizeof(T) == 4 || sizeof(T) == 8), void>;
+
+template <class T, class OidMapT, class OutIteratorT>
+constexpr floating_point_t<T> send(T value, const OidMapT&, OutIteratorT out) {
+    detail::write(value, out);
+}
+
+template <class OidMapT, class OutIteratorT, class CharT, class TraitsT, class AllocatorT>
+constexpr void send(const std::basic_string<CharT, TraitsT, AllocatorT>& value, const OidMapT&, OutIteratorT out) {
+    std::for_each(value.begin(), value.end(), [&] (auto v) { detail::write(v, out); });
```
thed636: There are similar places with this lambda. Is it better to make a function like

    template <typename Out>
    auto make_writer(Out out) { return [&] (auto v) { detail::write(v, out); };}

and then

    boost::for_each(value, detail::make_writer(out));

include/ozo/impl/query.h
```diff
@@ -0,0 +1,186 @@
...
+
+    template <class T>
+    static constexpr void write_meta(const oid_map_type& oid_map, std::size_t field, const std::reference_wrapper<T>& value, impl_type& result) {
+       write_meta<std::decay_t<T>>(oid_map, field, value.get(), result);
+    }
+
+    template <class T>
+    static constexpr is_nullable_t<T> write_meta(const oid_map_type& oid_map, std::size_t field, const T& value, impl_type& result) {
+       if (is_null(value)) {
+           write_typed_null_meta<std::decay_t<decltype(*value)>>(oid_map, field, result);
```
thed636: With **null_oid** and shorter signature it will look like this:

    write_null_meta(type_oid(oid_map,*value), field, result);

And it seems like the best way is to support **std::nullptr_t** and **std::nullopt_t** via type traits and define for them **is_built_in**  structure as **true_type**. In this case looks like you do not need specializations for **write_null_meta** and maybe even **write_null_meta** at all. E.g. you can overload

    send(std::nullptr_t, OidMap, Out) {}
    send(std::nullopt_t, OidMap, Out) {}

and **write_null_meta** is no more needed.

include/ozo/impl/query.h
```diff
@@ -0,0 +1,186 @@
...
+       }
+    }
+
+    template <class T>
+    static constexpr void write_meta(const oid_map_type& oid_map, std::size_t field, const std::weak_ptr<T>& value, impl_type& result) {
+       if (const auto locked = value.lock()) {
+           write_meta(oid_map, field, *locked, result);
+       } else {
+           write_typed_null_meta<T>(oid_map, field, result);
+       }
```
thed636: It looks like the shorter implementation should work well

    write_meta(oid_map, field, value.lock(), result);

Isn't it?

include/ozo/query.h
```diff
@@ -1,12 +1,37 @@
...
+struct is_query<T, std::void_t<
+    decltype(get_query_text(std::declval<const T&>())),
+    decltype(get_query_types(std::declval<const T&>())),
+    decltype(get_query_formats(std::declval<const T&>())),
+    decltype(get_query_lengths(std::declval<const T&>())),
+    decltype(get_query_values(std::declval<const T&>()))
+>> : std::true_type {};
+
+template <class ... ArgsT>
+auto make_query(const char* text, const ArgsT& ... args) {
```
thed636: Please correct me if I'm wrong, but in this case, const reference forbids us to move arguments, e.g. if I want to provide a string with no copy the const reference may be a problem.

include/ozo/binary_serialization.h
```diff
@@ -0,0 +1,138 @@
...
+    send(std::int32_t(1), oid_map, out);
+    std::for_each(value.begin(), value.end(),
+        [&] (const auto& v) {
+            send(detail::get_size(v), oid_map, out);
+            send(v, oid_map, out);
+        });
+}
+
+template <class OidMapT, class OutIteratorT, class AllocatorT>
+constexpr void send(const std::vector<char, AllocatorT>& value, const OidMapT&, OutIteratorT out) {
```
thed636: Why is this specialization needed for std::vector<char>?

include/ozo/impl/query.h
```diff
@@ -0,0 +1,186 @@
...
+        result.types[field] = type_oid<T>(oid_map);
+        result.formats[field] = 1;
+        result.lengths[field] = 0;
+    }
+
+    static constexpr void write_null_meta(const oid_map_type&, std::size_t field, impl_type& result) noexcept {
+        result.types[field] = 0;
+        result.formats[field] = 1;
+        result.lengths[field] = 0;
+    }
```
thed636: Is it better to use single function but with different signature

    static constexpr void write_null_meta(oid_t oid, std::size_t field, impl_type& result) noexcept {
         result.types[field] = oid;
         result.formats[field] = 1;
         result.lengths[field] = 0;
     }

So you can use **null_oid** constant which is defined in **type_traits.h** instead of different functions.

thed636: IMO it is better to use proxy for result, something like this:

    struct type_meta_proxy {
        std::size_t index;
        impl_type& result;
    
        auto& type() { return result.types[index]; }
        auto& format() { return result.formats[index]; }
        auto& length() { return result.lengths[index];}
        auto& buffer() { return result.buffer; }
    };

    static constexpr void write_null_meta(oid_t oid, type_meta_proxy result) noexcept {
         result.type() = oid;
         result.format() = 1; // text or binary? const is needed
         result.length() = 0;
     }

include/ozo/binary_serialization.h
elsid: Why should I do this? For what reason we need to increase compexity of code to write simple sequence of functions calls? I don't see any profit.

include/ozo/impl/query.h
elsid: It is an interface for direct use to call libpq functions without any extra conversions. Please, provide some basic example, how you want to see this interface with integration with libpq.

include/ozo/query.h
elsid: We anyway need to serialize it into binary representation for lipbq and will copy. If you want optimization when you already have binary serialized value and just want to reference it, then you will need more complex implementation for query constructor. Do we need it now?

include/ozo/query_builder.h
elsid: Ok, I will remove this.

include/ozo/binary_serialization.h
thed636: Well, if you need to change a function signature or something like this you will do it once. And more generally, if you need to write different headers for different types you can use traits instead of different functions.

BTW, I was wrong about boilerplate code.
http://www.boost.org/doc/libs/1_61_0/libs/hana/doc/html/group__group-Struct.html#gad301dd8e9fb4639d7874619c97d6d427
hana::members can return a tuple of values of an adapted structure. In this case, if you define the header structure it will be easy to understand what the header is.

include/ozo/impl/query.h
elsid: Done

include/ozo/impl/query.h
elsid: Done

include/ozo/binary_serialization.h
elsid: Done

include/ozo/binary_serialization.h
elsid: Ok, let's use it.

include/ozo/binary_serialization.h
elsid: Because we serialize it as `bytea` type.

include/ozo/binary_serialization.h
elsid: Ok, but capture iterator by reference is bad idea.

include/apq/query.h
elsid: Done

include/ozo/impl/query.h
elsid: But we still need to do different logic with nullable and not types? What is the point to move it to the send functions?

include/ozo/binary_serialization.h
elsid: Done

thed636 (approved):

PR #7, opened 2017-12-26
Add connection implementation.

thed636:
Connection is a category which is defined in apq/connection.h
To obtain connection ConnectionProvider must exists. ConnectionProvider
category is defined in apq/connection.h too. Connection is
ConnectionProvider itself. To obtain a connection from connection
provider get_connection() function exists.
Connectable is a category of objects which contains Connection.
Connection is Connectable itself. Wrapped connection is Connectable
if wrapper is ConnectionWrapper.

The basic implementation of the Connection category can be found
in apq/impl/connection.h.

Calss connection_info is a ConnectionProvider which creates connection
via connection info string. To perform the connection async_connect
operation is defined.

### DIFF
--- a/include/apq/async_result.h
+++ b/include/apq/async_result.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <boost/asio/async_result.hpp>
+#include <utility>
+#include <apq/asio.h>
+
+namespace libapq {
+
+//We can reuse boost::asio::async_result since it is not binded to error code type
+using asio::async_result;
+using asio::handler_type;
+
+/**
+* This is a part of compatibility layer with Boos.Asio 1.66
+*/
+template <typename CompletionToken, typename Signature>
+struct async_completion {
+    explicit async_completion(CompletionToken& token)
+    : completion_handler(std::move(token)),
+      result(completion_handler) {
+    }
+
+    using completion_handler_type = typename libapq::handler_type<CompletionToken, Signature>::type;
+
+    completion_handler_type completion_handler;
+    async_result<completion_handler_type> result;
+};
+
+} // namespace libapq
--- a/include/apq/connection.h
+++ b/include/apq/connection.h
@@ -1,10 +1,376 @@
 #pragma once
+#include <iostream>
+
+#include <apq/async_result.h>
+#include <apq/error.h>
+#include <apq/type_traits.h>
+#include <apq/asio.h>
+#include <apq/concept.h>
+
+#include <apq/detail/bind.h>
+#include <apq/impl/connection.h>
 
 namespace libapq {
 
-class connection {
-public:
-    connection() {}
+using impl::pg_conn_handle;
+
+using no_statistics = decltype(hana::make_map());
+
+/**
+* Marker to tag a connection type.
+*/
+template <typename, typename = std::void_t<>>
+struct is_connection : std::false_type {};
+
+/**
+* We define the connection to database not as a concrete class or type,
+* but as an entity which must support some traits. The reason why we
+* choose such strategy is - we want to provide flexibility and extension ability
+* for implementation and testing. User can implemet a connection bound additional
+* functionality.
+*
+* This is how connection is determined. Connection is an object for which these
+* next free functions are applicable:
+*
+*   * get_connection_oid_map()
+*     Must return reference or proxy for connection's oid map object, which allows
+*     to read and modify it. Object must be created via register_types<>() template
+*     function or be empty_oid_map in case if no custom types are used with the
+*     connection.
+*
+*   * get_connection_socket()
+*     Must return reference or proxy for socket io stream object which allows to bind
+*     the connection to the asio io_context, currently boost::asio::posix::stream_descriptor
+*     is suppurted only.
+*
+*   * get_connection_handle()
+*     Must return reference or proxy for pg_conn_handle object
+*
+*   * get_connection_error_context()
+*     Must return reference or proxy for additional error context. There is no
+*     mechanism to privide context depent information via [std|boost]::error_code. So
+*     we decide to use connection for this purpose since native libpq error message is
+*     bound to connection and all asyncronous operation of the library is bound to
+*     connection too. Currently std::string supported as a context object.
+*/
+template <typename T>
+struct is_connection<T, std::void_t<
+    decltype(get_connection_oid_map(std::declval<T&>())),
+    decltype(get_connection_socket(std::declval<T&>())),
+    decltype(get_connection_handle(std::declval<T&>())),
+    decltype(get_connection_error_context(std::declval<T&>())),
+    decltype(get_connection_oid_map(std::declval<const T&>())),
+    decltype(get_connection_socket(std::declval<const T&>())),
+    decltype(get_connection_handle(std::declval<const T&>())),
+    decltype(get_connection_error_context(std::declval<const T&>()))
+>> : std::true_type {};
+
+
+template <typename, typename = std::void_t<>>
+struct is_connection_wrapper : std::false_type {};
+
+template <typename T>
+struct is_connection_wrapper<::std::shared_ptr<T>> : std::true_type {};
+template <typename T, typename D>
+struct is_connection_wrapper<::std::unique_ptr<T, D>> : std::true_type {};
+template <typename T>
+struct is_connection_wrapper<::std::optional<T>> : std::true_type {};
+template <typename T>
+struct is_connection_wrapper<::boost::shared_ptr<T>> : std::true_type {};
+template <typename T>
+struct is_connection_wrapper<::boost::scoped_ptr<T>> : std::true_type {};
+template <typename T>
+struct is_connection_wrapper<::boost::optional<T>> : std::true_type {};
+
+template <typename T>
+constexpr auto ConnectionWrapper = is_connection_wrapper<std::decay_t<T>>::value;
+
+template <typename T>
+constexpr auto Connection = is_connection<std::decay_t<T>>::value;
+
+/**
+* Connection type traits
+*/
+template <typename T>
+struct connection_traits {
+    using oid_map = std::decay_t<decltype(get_connection_oid_map(std::declval<T&>()))>;
+    using socket = std::decay_t<decltype(get_connection_socket(std::declval<T&>()))>;
+    using handle = std::decay_t<decltype(get_connection_handle(std::declval<T&>()))>;
+    using error_context = std::decay_t<decltype(get_connection_error_context(std::declval<T&>()))>;
 };
 
-} // namespace apq
+/**
+* Connection public access functions section 
+*/
+
+/**
+* Overloaded version of unwrap function for Connection.
+* Returns connection object itself.
+*/
+template <typename T>
+inline decltype(auto) unwrap_connection(T&& conn, 
+        Require<!ConnectionWrapper<T>>* = 0) noexcept {
+    return std::forward<T>(conn);
+}
+
+/**
+* Unwraps wrapped connection recusively.
+* Returns unwrapped connection object.
+*/
+template <typename T>
+inline decltype(auto) unwrap_connection(T&& conn, 
+        Require<ConnectionWrapper<T>>* = 0) noexcept {
+    return unwrap_connection(*conn);
+}
+
+template <typename, typename = std::void_t<>>
+struct is_connectiable : std::false_type {};
+
+template <typename T>
+struct is_connectiable <T, std::void_t<
+    decltype(unwrap_connection(std::declval<T&>()))
+>>: std::integral_constant<bool, Connection<
+    decltype(unwrap_connection(std::declval<T&>()))
+>> {};
+
+template <typename T>
+constexpr auto Connectable = is_connectiable<std::decay_t<T>>::value;
+
+/**
+* Function to get connection traits type from object
+*/
+template <typename T, typename = Require<Connectable<T>>>
+constexpr connection_traits<std::decay_t<decltype(unwrap_connection(std::declval<T>()))>>
+    get_connection_traits(T&&) noexcept {return {};}
+
+/**
+* Returns PostgreSQL connection handle of the connection.
+*/
+template <typename T, typename = Require<Connectable<T>>>
+inline decltype(auto) get_handle(T&& conn) noexcept {
+    using impl::get_connection_handle;
+    return get_connection_handle(
+        unwrap_connection(std::forward<T>(conn)));
+}
+
+/**
+* Returns PostgreSQL native connection handle of the connection.
+*/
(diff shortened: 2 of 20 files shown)


include/apq/async_result.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <boost/asio/async_result.hpp>
+#include <utility>
+#include <apq/asio.h>
+
+namespace libapq {
+
+//We can reuse boost::asio::async_resul since it is not binded to error code type
```
elsid: boost::asio::async_resul**t**

include/apq/connection.h
```diff
@@ -1,10 +1,314 @@
...
+/**
+* Connection public access functions section 
+*/
+
+/**
+* Overloaded version of unwrap function for Connection.
+* Returns connection object itself.
+*/
+template <typename T>
+inline decltype(auto) unwrap_connection(T&& conn,
```
elsid, line 111: Why inline modifier is required here?

include/apq/connection_info.h
```diff
@@ -0,0 +1,71 @@
...
+
+    using connection = impl::connection<OidMap, Statistics>;
+
+public:
+    using connectiable_type = std::shared_ptr<connection>;
+
+    connection_info(io_context& io, std::string conn_str, Statistics statistics = Statistics{})
+    : io_(io), raw_(std::move(conn_str)), statistics_(statistics) {}
+
+    std::string to_string() const {return raw_;}
```
elsid: Copy string here is not so good idea. Return const reference or string_view.

include/apq/connection_info.h
```diff
@@ -0,0 +1,71 @@
...
+}
+
+} // namespace detail
+
+template <
+    typename OidMap = empty_oid_map, 
+    typename Statistics = no_statistics>
+class connection_info {
+    io_context& io_;
+    std::string raw_;
```
elsid: Does connection_info require to own connection info string? Why don't just use string_view?

include/apq/impl/async_connect.h
```diff
@@ -0,0 +1,172 @@
...
+    int fd = PQsocket(get_native_handle(conn));
+    if (fd == -1) {
+        return make_error_code(error::pq_socket_failed);
+    }
+
+    int new_fd = dup(fd);
+    if (new_fd == -1) {
+        using namespace boost::system;
+        return make_error_code(error_code{errno, get_posix_category()},
+            "dup(fd)");
```
elsid: ```"dup(fd)``` is not very informative message. Use ```std::strerror(errno)```.

include/apq/impl/async_connect.h
```diff
@@ -0,0 +1,172 @@
...
+    Handler handler_;
+
+    void done(error_code ec = error_code{}) {
+        get_io_context(conn_)
+            .post(detail::bind(std::move(handler_), std::move(ec)));
+    }
+
+    void perform(const std::string& conninfo) {
+        error_code ec = start_connection(conn_, conninfo);
+        if (ec) {
```
elsid: Allow to reuse ec variable is not so good idea. Why just not ```if (const auto ec = start_connection(conn_, conninfo)) {```?

include/apq/impl/async_connect.h
```diff
@@ -0,0 +1,172 @@
...
+            return done();
+
+        case PGRES_POLLING_WRITING:
+            return write_poll(conn_, std::move(*this));
+
+        case PGRES_POLLING_READING:
+            return read_poll(conn_, std::move(*this));
+
+        case PGRES_POLLING_FAILED:
+        default:
```
elsid: Why do you ignore other values?

include/apq/impl/connection.h
```diff
@@ -0,0 +1,102 @@
...
+inline error_code rebind_connection_io_context(Connection& conn, io_context& io) {
+    if (std::addressof(get_io_service(conn)) != std::addressof(io)) {
+        asio::posix::stream_descriptor s{io};
+        error_code ec;
+        decltype(auto) socket = get_socket(conn);
+        s.assign(socket.native_handle(), ec);
+        if (ec) {
+            return ec;
+        }
+        socket.release();
```
elsid, line 111: Why here is explicit release? Isn't only assign will be enough?

include/apq/impl/connection.h
```diff
@@ -0,0 +1,102 @@
...
+
+template <typename ...Ts>
+inline auto& get_connection_statistics(
+        connection<Ts...>& ctx) noexcept {
+    return ctx.statistics_;
+}
+
+inline std::string connection_error_message(pg_native_handle_type handle) {
+    const char* msg = PQerrorMessage(handle);
+    return msg ? boost::trim_right_copy(std::string(msg)) : std::string{};
```
elsid: May be just use string_view for return type?

elsid (changes requested):

include/apq/impl/connection.h
thed636: Because release don't close fd :) socket is RAII object. We need to transfer fd from one object to another.

include/apq/impl/connection.h
thed636: Who owns the const char* msg? What is a lifetime of the msg?

include/apq/impl/async_connect.h
thed636: They are not ignored - we didn't expect them at all. So any other value is unexpected and leads to the end of the procedure with an error. Of course, we can remove PGRES_POLLING_FAILED and nothing will change now. Or we can add one more error code for PGRES_POLLING_FAILED only.

include/apq/impl/async_connect.h
thed636: Well, std::strerror(errno) is made inside posix_category::message, so "dup(fd)" gives an information about where the error occurred. The message can be changed to contain more words if it really needed.

include/apq/connection_info.h
thed636: In case of string_view usage, our user will be forced to store this string somewhere else and have a lot of fun with string lifetime. This object is a connection producer, so it is not expected to create/copy it often. I do not see any good reason to perform such optimisation here. Do you?

include/apq/connection.h
thed636: Standard practice to avoid duplicate specialized instantiation in different translation units. Any problems?

include/apq/async_result.h
thed636: **Done**

include/apq/impl/async_connect.h
thed636: **Done**

include/apq/impl/async_connect.h
thed636: **Done**

include/apq/impl/connection.h
thed636: **Done**

include/apq/connection.h
elsid: ok

include/apq/impl/connection.h
elsid: Ok, I see, missed that.

include/apq/connection_info.h
elsid: I don't, ok.

include/apq/impl/async_connect.h
elsid: ok

include/apq/impl/async_connect.h
```diff
@@ -0,0 +1,164 @@
...
+inline error_code pq_assign_socket(T& conn) {
+    int fd = PQsocket(get_native_handle(conn));
+    if (fd == -1) {
+        return error::pq_socket_failed;
+    }
+
+    int new_fd = dup(fd);
+    if (new_fd == -1) {
+        set_error_context(conn, "error while dup(fd) for socket stream");
+        return error_code{errno, boost::system::get_posix_category()};
```
elsid: Looks like `boost::system::get_posix_category` is [deprecated](https://github.com/boostorg/system/blob/41e0404847aef1e7f8f987c8a4c5cbca68b453fd/include/boost/system/error_code.hpp#L204).

include/apq/connection.h
```diff
@@ -1,10 +1,375 @@
...
+    using impl::get_connection_error_context;
+    get_connection_error_context(unwrap_connection(conn)) = std::forward<Ctx>(ctx);
+}
+
+/**
+* Reset the connection error context.
+*/
+template <typename T>
+inline Require<Connectiable<T>> reset_error_context(T& conn) {
+    set_error_context(conn, "");
```
dpotepalov: why ""? Looks strange, since Ctx is a template in setter and getter. Is this function really necessary? It's signature does not allow deduction of Ctx, maybe just use `set_error_context(conn, Ctx{})` wherever necessary?

include/apq/error.h
```diff
@@ -0,0 +1,78 @@
...
+                return "pq_connection_start_failed - PQConnectionStart function failed";
+            case pq_socket_failed:
+                return "pq_socket_failed - PQSocket returned -1 as fd - it seems like there is no connection";
+            case pq_connection_status_bad:
+                return "pq_connection_status_bad - PQstatus returned CONNECTION_BAD";
+            case pq_connect_poll_failed:
+                return "pq_connect_poll_failed - PQconnectPoll function failed";
+        }
+        std::ostringstream error;
+        error << "no message for value: " << value;
```
dpotepalov: seems excessive, a simple `return "no message for value: " + std::to_string(value);` should suffice

elsid:
We need to figure out, why tests don't pass.
```
[ SHOULD   ] start connection, assign socket and wait for write complete
pure virtual method called
terminate called without an active exception

Program received signal SIGABRT, Aborted.
0x00007ffff606c860 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff606c860 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff606dec9 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff69e0d57 in __gnu_cxx::__verbose_terminate_handler () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x00007ffff69de8c6 in __cxxabiv1::__terminate (handler=<optimized out>) at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:47
#4  0x00007ffff69de913 in std::terminate () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:57
#5  0x00007ffff69df721 in __cxxabiv1::__cxa_pure_virtual () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/pure.cc:50
#6  0x0000555555649f44 in (anonymous namespace)::pq_start_connection (c=..., conninfo="conninfo") at /home/elsid/dev/apq2/tests/async_connect.cpp:89
#7  libapq::impl::start_connection<std::shared_ptr<(anonymous namespace)::connection<> > > (conn=..., conninfo="conninfo") at /home/elsid/dev/apq2/include/apq/impl/async_connect.h:65
#8  libapq::impl::async_connect_op<libapq::testing::callback_handler<libapq::testing::callback_mock<> >, std::shared_ptr<(anonymous namespace)::connection<boost::hana::detail::map_impl<boost::hana::detail::hash_table<>, boost::hana::basic_tuple<> > > > >::perform (this=<optimized out>, conninfo="conninfo") at /home/elsid/dev/apq2/include/apq/impl/async_connect.h:109
#9  libapq::impl::async_connect<std::shared_ptr<(anonymous namespace)::connection<> >, libapq::testing::callback_handler<libapq::testing::callback_mock<> > > (conninfo="conninfo", conn=std::shared_ptr (count 1, weak 0) 0x55555593ba20, 
    handler=...) at /home/elsid/dev/apq2/include/apq/impl/async_connect.h:158
#10 (anonymous namespace)::GTEST<testing::v1::detail::string<'\"', 'l', 'i', 'b', 'a', 'p', 'q', ':', ':', 'a', 's', 'y', 'n', 'c', '_', 'c', 'o', 'n', 'n', 'e', 'c', 't', '(', ')', '\"', '\000'>, testing::v1::detail::string<> >::TestBodyImpl(testing::v1::detail::TestRun &) (tr_gtest=..., this=0x7fffffffe1e0) at /home/elsid/dev/apq2/tests/async_connect.cpp:132
#11 0x0000555555644ae2 in (anonymous namespace)::GTEST<testing::v1::detail::string<'\"', 'l', 'i', 'b', 'a', 'p', 'q', ':', ':', 'a', 's', 'y', 'n', 'c', '_', 'c', 'o', 'n', 'n', 'e', 'c', 't', '(', ')', '\"', '\000'>, testing::v1::detail::string<> >::TestBody(void) (this=<optimized out>) at /home/elsid/dev/apq2/tests/async_connect.cpp:119
#12 0x00005555556b1751 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#13 0x00005555556ab7cb in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#14 0x000055555568f4ce in testing::Test::Run() ()
#15 0x000055555568fe04 in testing::TestInfo::Run() ()
#16 0x0000555555690486 in testing::TestCase::Run() ()
#17 0x0000555555697352 in testing::internal::UnitTestImpl::RunAllTests() ()
#18 0x00005555556b2917 in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#19 0x00005555556ac607 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#20 0x0000555555695f16 in testing::UnitTest::Run() ()
#21 0x00005555556d1780 in RUN_ALL_TESTS() ()
#22 0x00005555556d170f in main ()
```

thed636:
@elsid thanks for the fix with this vtable issue, I've rebased on it and fixed typo Connectiable -> Connectable.

include/apq/impl/async_connect.h
thed636: **Done**

include/apq/error.h
thed636: **Done**

include/apq/connection.h
thed636: **Done**
I've changed it to type deduction from get_error_context() result
In this case, a user can override reset_error_context() for his own connection type if it needed.

elsid (approved):

PR #17, opened 2018-03-18
Result

dpotepalov:


include/ozo/impl/row.h
```diff
@@ -0,0 +1,105 @@
...
+
+template <typename T>
+constexpr auto IsFusionSequence = boost::fusion::traits::is_sequence<T>::value;
+
+template <typename RowData, typename OidMap, typename Row>
+Require<Iterable<RowData> && IsFusionSequence<Row>, error_code>
+convert_row(RowData&& row_data, const OidMap& oid_map, Row& row) {
+    error_code ec = error::ok;
+    auto data_it = begin(row_data);
+    auto data_end = end(row_data);
```
thed636: I think you can make the dimension check here instead of checks in for_each. The Row must be a Sequence, so you can get a size of the sequence and compare it with std::distance between data_it and data_end.

include/ozo/binary_deserialization.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <ozo/error.h>
+#include <ozo/type_traits.h>
+#include <ozo/concept.h>
+#include <ozo/detail/endian.h>
+
+namespace ozo {
+
+inline error_code recv(oid_t, const char* bytes, std::size_t size, std::string& value) {
+    // TODO: client to server encoding conversion?
```
thed636: What do you mean?

include/ozo/impl/result.h
```diff
@@ -0,0 +1,49 @@
...
+#include <ozo/error.h>
+#include <ozo/type_traits.h>
+#include <ozo/concept.h>
+#include <ozo/impl/row.h>
+
+namespace ozo::impl {
+
+template <typename Rows, typename OidMap, typename Iterator>
+Require<Iterable<Rows> && (ForwardIterator<Iterator> || InsertIterator<Iterator>), error_code>
+convert_result(Rows&& rows, const OidMap& oid_map, Iterator result) {
```
thed636: IMO it is better to use ```out``` name for output result to avoid using a part of function name as a parameter name.

include/ozo/impl/result.h
```diff
@@ -0,0 +1,49 @@
...
+    error_code ec = error::ok;
+    for (auto row : rows) {
+        if constexpr (ForwardIterator<Iterator>) {
+            ec = convert_row(std::move(row), oid_map, *result++);
+        } else if constexpr (InsertIterator<Iterator>) {
+            using row_value_t = typename Iterator::container_type::value_type;
+
+            auto result_row = row_value_t{};
```
thed636: I think it is better to use an overloaded function instead of ```constexpr``` because even with ForwardIterator and not a default constructible type you still can get here a compile error here.

include/ozo/impl/result.h
```diff
@@ -0,0 +1,49 @@
...
+}
+
+template <typename Rows>
+class result {
+    using rows_t = std::decay_t<Rows>;
+public:
+    result(rows_t&& rows) : rows_(std::move(rows)) {}
+
+    template <typename Iterator, typename OidMap>
+    error_code convert(Iterator&& result, const OidMap& oid_map) {
```
thed636: IMO ```in``` or something like this is better than name ```result``` which is also a name of the class.

include/ozo/binary_deserialization.h
```diff
@@ -0,0 +1,36 @@
...
+        return error::integer_value_size_mismatch;
+    }
+
+    const T* p_be = reinterpret_cast<const T*>(bytes);
+    value = detail::convert_from_big_endian(*p_be);
+    return error::ok;
+}
+
+template <typename OidMap, typename T>
+inline error_code convert(oid_t oid, const char* bytes, std::size_t size, const OidMap& oid_map, T& value) {
```
thed636: IMO it is better to use ```value``` type or it's concept instead of these three parameters here:
```c++
oid_t oid, const char* bytes, std::size_t size
```

include/ozo/libpq.h
```diff
@@ -0,0 +1,143 @@
...
+            } else {
+                *this = {};
+            }
+        }
+
+        PGresult* r_ {nullptr};
+        int row_ {0};
+        int col_ {0};
+
+        value v_ {nullptr, 0, 0};
```
thed636: The value contains same data elements as iterator itself - it is very strange and redundant.

include/ozo/libpq.h
```diff
@@ -0,0 +1,143 @@
...
+                v_ = {r_, row_};
+            } else {
+                *this = {};
+            }
+        }
+
+        PGresult * r_;
+        int row_;
+
+        row v_ {nullptr, 0};
```
thed636: Same as above.

include/ozo/libpq.h
```diff
@@ -0,0 +1,143 @@
...
+};
+
+
+class result
+{
+public:
+    class iterator : public boost::iterator_facade<
+        iterator,
+        row,
+        boost::forward_traversal_tag, // can be random_access, but we do not need it
```
thed636: Why not?

include/ozo/libpq.h
```diff
@@ -0,0 +1,143 @@
...
+
+    row() = default;
+    row(PGresult* r, int row) : r_(r), row_(row) {}
+
+    iterator begin() const { return iterator{r_, row_, 0}; }
+    iterator end() const { return iterator{}; }
+
+    // TODO: replace usage of these in impl::row with a random access row iterator
+    std::size_t size() { return PQnfields(r_); }
+    value at(int i) { return value{r_, row_, i}; }
```
thed636: What happened here in case of ``` i``` is out of range?

tests/row.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+struct fusion_adapted {
+    mock_value a;
+    mock_value_2 b;
+};
+
+// TODO: maybe we should try to hide this behind OZO_ADAPT_STRUCT or something?
+// In order to hide usage of fusion or hana as an implementation detail.
+// Would involve a bunch of generated code, though.
+BOOST_FUSION_ADAPT_STRUCT(fusion_adapted, a, b)
```
thed636: We shouldn't. It is not the best idea to force a user to use yet another one adaptation system for types and structures. The good choice is to support these two systems simultaneously and ask a user only to provide additional information like type name in a target database.

include/ozo/impl/result.h
```diff
@@ -0,0 +1,49 @@
...
+#include <ozo/type_traits.h>
+#include <ozo/concept.h>
+#include <ozo/impl/row.h>
+
+namespace ozo::impl {
+
+template <typename Rows, typename OidMap, typename Iterator>
+Require<Iterable<Rows> && (ForwardIterator<Iterator> || InsertIterator<Iterator>), error_code>
+convert_result(Rows&& rows, const OidMap& oid_map, Iterator result) {
+    error_code ec = error::ok;
```
thed636: A right variable indicating no error will be:

```c++
error_code ec;
```

Since ```error::ok``` is still an error code. It is just reserved value to not be confused with zero value error code.

tests/binary_deserialization.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    SHOULD("[convert INT2OID to int16_t]") {
+        const int16_t expected = 7;
+        int16_t got;
+
+        const int16_t bytes_storage = detail::convert_to_big_endian(expected);
+        const char* bytes = reinterpret_cast<const char*>(&bytes_storage);
+        const auto size = sizeof(int16_t);
+
+        EXPECT_EQ(ok, convert(INT2OID, bytes, size, oid_map, got));
```
thed636: You should expect 
```c++
!convert(INT2OID, bytes, size, oid_map, got)
```
here.

thed636 (dismissed):

include/ozo/binary_deserialization.h
dpotepalov: I misinterpreted text's recv() code in postgres, it also does encoding conversions on strings, so I thought that maybe we should do the same. We should not, it's done by the server. I'll remove the comment.

tests/binary_deserialization.cpp
```diff
@@ -0,0 +1,59 @@
...
+        int x;
+        EXPECT_EQ(oid_type_mismatch, convert(TEXTOID, nullptr, 0, oid_map, x));
+    }
+
+    SHOULD("[convert INT2OID to int16_t]") {
+        const int16_t expected = 7;
+        int16_t got;
+
+        const int16_t bytes_storage = detail::convert_to_big_endian(expected);
+        const char* bytes = reinterpret_cast<const char*>(&bytes_storage);
```
elsid: Why perform data conversions in the test? Just define value as it should be.

elsid (changes requested):

include/ozo/impl/row.h
dpotepalov: done

include/ozo/impl/result.h
dpotepalov: done

include/ozo/impl/result.h
dpotepalov: done

include/ozo/impl/result.h
dpotepalov: done

include/ozo/binary_deserialization.h
dpotepalov: done

include/ozo/libpq.h
dpotepalov: done

include/ozo/libpq.h
dpotepalov: done

include/ozo/libpq.h
dpotepalov: it's unnecessary for conversion as implemented now, because generic detail::result iterates forwards through libpq::result rows. If we decide to make detail::result (and thus ozo::result) a full-fledged container with random access, then we'll have to implement random access here.

include/ozo/libpq.h
dpotepalov: done - mimicked std:: containers behaviour

tests/row.cpp
dpotepalov: ok, comment removed

include/ozo/impl/result.h
dpotepalov: done

tests/binary_deserialization.cpp
dpotepalov: done

tests/binary_deserialization.cpp
dpotepalov: done

include/ozo/libpq.h
```diff
@@ -0,0 +1,159 @@
...
+#include <libpq-fe.h>
+
+#include <boost/iterator/iterator_facade.hpp>
+#include <memory>
+#include <vector>
+
+
+namespace ozo::libpq {
+
+class value {
```
thed636: IMO no friend class for value is needed at all. Did you try something like proxy object?
```c++
class value {
public:
    struct data { PGresult* res; int row; int col; };
    value(const data& v) : v_(v) {}
//...
private:
    data& v_;
};

//...

class row {
public:
    class iterator : public boost::iterator_facade<
//...
        iterator,
        value,
        boost::forward_traversal_tag, // maybe will implement random access later
        value, // not a reference, but proxy
        int
    > {
        public:
        //...
        private:
            value::data v_;
            value dereference() const {
                return {v_};
            }
//...
```

include/ozo/libpq.h
```diff
@@ -0,0 +1,159 @@
...
+    friend class row;
+};
+
+
+class row {
+public:
+    class iterator : public boost::iterator_facade<
+        iterator,
+        value,
+        boost::forward_traversal_tag, // maybe will implement random access later
```
thed636: Why not implement it right now? Any problems?

include/ozo/libpq.h
thed636: IMO there are not so many changes is needed to make a random access iterator

include/ozo/libpq.h
```diff
@@ -0,0 +1,159 @@
...
+
+private:
+    unique_result_ptr r_;
+};
+
+inline row::iterator begin(const row& r) { return r.begin(); }
+inline row::iterator end(const row& r) { return r.end(); }
+
+inline result::iterator begin(const result& r) { return r.begin(); }
+inline result::iterator end(const result& r) { return r.end(); }
```
thed636: Are these implementations really needed?

include/ozo/libpq.h
dpotepalov: I considered such an option and decided that creating an object inside a dereference() method looks a little awkward. If you both decide that it doesn't - ok, proxy it is. There is no need for value::data, actually - you can store res, row and col in row as separate variables and create an instance of value in dereference()

include/ozo/libpq.h
dpotepalov: Unused code? Three more methods of iterator_facade need to be implemented. Not much work, sure, but that code will be just sitting there doing nothing, potentially requiring maintenance in the future.

include/ozo/libpq.h
dpotepalov: actually not, they were needed at some point. I'll remove them.

include/ozo/libpq.h
thed636: --Well, IMO return wrapped reference is better than copy pointer + two integers.--

I'm wrong. The wrapped reference to data is bad way - the only right version is to copy all coordinates into proxy

include/ozo/binary_deserialization.h
```diff
@@ -0,0 +1,157 @@
...
+            )
+        );
+    }
+};
+
+template <typename T, typename M, typename Out>
+inline void recv(const value<T>& in, const oid_map_t<M>& oid_map, Out& out) {
+
+    if (!accepts_oid(oid_map, out, in.oid())) {
+        throw system_error(error::oid_type_mismatch, "unexpected oid "
```
dpotepalov: why throw? I thought we decided to use error codes. Although this kind of error does seem unrecoverable.

include/ozo/binary_deserialization.h
```diff
@@ -0,0 +1,157 @@
...
+recv_result(const basic_result<T>& in, const oid_map_t<M>& oid_map, Out out) {
+    for (auto row : in) {
+        typename Out::container_type::value_type v{};
+        recv_row(row, oid_map, v);
+        *out++ = std::move(v);
+    }
+}
+
+template <typename T, typename M, typename Out>
+Require<InsertIterator<Out>>
```
dpotepalov: seems unnecessary, since Out is just basic_result<T> and body does not require InsertIterator support

include/ozo/impl/io.h
```diff
@@ -7,6 +7,17 @@
 
 namespace ozo {
 namespace impl {
+
+using result_ptr = std::unique_ptr<PGresult, decltype(&PQclear)>;
+
+/**
+* Here we describe constants for query result format
```
dpotepalov: enum name seems self-explanatory, IMO comment is unnecessary here

tests/binary_deserialization.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+using namespace ::testing;
+using ::ozo::testing::pg_result_mock;
+
+GTEST("ozo::recv") {
+    auto oid_map = ozo::empty_oid_map{};
+    StrictGMock<pg_result_mock> mock{};
+    ::ozo::value<pg_result_mock> value({object(mock), 0, 0});
+
+    SHOULD("return type mismatch error if oid does not match the type") {
```
dpotepalov: throw type mismatch error?

tests/binary_deserialization.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+
+GTEST("ozo::recv_row") {
+    auto oid_map = ozo::empty_oid_map{};
+    StrictGMock<pg_result_mock> mock{};
+    ::ozo::row<pg_result_mock> row({object(mock), 0, 0});
+
+    SHOULD("return type mismatch error if size of tuple does not equal to row size") {
```
dpotepalov: as above - throw type mismatch error?

include/ozo/binary_deserialization.h
thed636: Well, the error_code and context depended information will be provided for user callbacks by means of internal logic and it's arguments. On this level the error_code is not useful because it can not provide context depended information about an error.

include/ozo/binary_deserialization.h
thed636: Yep, thanks, it seems like I forgot a unit test for the case.

include/ozo/impl/io.h
thed636: Yep, you right

tests/binary_deserialization.cpp
thed636: Yep, thanks.

tests/binary_deserialization.cpp
thed636: Yep, thanks a lot!

include/ozo/binary_deserialization.h
thed636: The comment is moved to the different part of code by github - funny :)

elsid (approved):

PR #20, opened 2018-03-20
Query conf

elsid:


include/ozo/query_conf.h
```diff
@@ -0,0 +1,487 @@
...
+}
+
+template <class ... QueriesT>
+void check_for_duplicates(const hana::tuple<QueriesT ...>& queries) {
+    hana::fold_left(
+        queries, hana::make_set(),
+        [] (auto set, auto query) {
+            const auto result = hana::insert(set, query.name);
+            if (set == result) {
+                throw std::invalid_argument(hana::to<const char*>("Duplicate declaration for query: "_s + query.name));
```
thed636, line 276: Why hana::string ?

elsid: `query.name` is also hana::string, so we can concat them in compile time.

thed636 (approved):

PR #40, opened 2018-05-23
Add tests for custom structs serialization

elsid:


thed636:
This PR is no more needed since adapted structures support was introduced.

PR #43, opened 2018-06-01
Pool fixes

thed636:
Fix connection pool regarding to benchmark branch, but a little bit cleaner. Remove a lot of redundant staff.

### DIFF
--- a/include/ozo/connection_pool.h
+++ b/include/ozo/connection_pool.h
@@ -8,59 +8,88 @@
 namespace ozo {
 namespace impl {
 
-template <typename ...Ts>
-using connection_pool = yamail::resource_pool::async::pool<connection<Ts...>>;
+template <typename Provider>
+struct get_connection_pool {
+    using type = yamail::resource_pool::async::pool<
+        std::decay_t<decltype(unwrap_connection(std::declval<connectable_type<Provider>&>()))>
+    >;
+};
 
-template <typename ...Ts>
+template <typename Provider>
+using connection_pool = typename get_connection_pool<Provider>::type;
+
+template <typename Provider>
 struct pooled_connection {
-    using handle_type = typename connection_pool<Ts...>::handle;
+    using handle_type = typename connection_pool<Provider>::handle;
+    using underlying_type = typename handle_type::value_type;
 
     handle_type handle_;
 
     pooled_connection(handle_type&& handle) : handle_(std::move(handle)) {}
 
-    decltype(auto) empty() const {return handle_.empty();}
+    bool empty() const {return handle_.empty();}
 
-    void reset(typename handle_type::value_type&& v) {
-        reset(std::move(v));
+    void reset(underlying_type&& v) {
+        handle_.reset(std::move(v));
     }
 
     ~pooled_connection() {
-        if (connection_bad(*this)) {
+        if (!empty() && connection_bad(*this)) {
             handle_.waste();
         }
     }
 
     friend auto& unwrap_connection(pooled_connection& conn) noexcept {
-        return *(conn.handle_);
+        return unwrap_connection(*(conn.handle_));
     }
     friend const auto& unwrap_connection(const pooled_connection& conn) noexcept {
-        return *(conn.handle_);
+        return unwrap_connection(*(conn.handle_));
     }
 };
-template <typename ...Ts>
-using pooled_connection_ptr = std::shared_ptr<pooled_connection<Ts...>>;
+template <typename Provider>
+using pooled_connection_ptr = std::shared_ptr<pooled_connection<Provider>>;
 
-template <typename Connector, typename OidMap, typename Statistics, typename Handler>
+template <typename Provider, typename Handler>
 struct pooled_connection_wrapper {
-    io_context& io_;
-    Connector connector_;
-    Statistics statistics_;
+    Provider provider_;
     Handler handler_;
+
+    using connection = pooled_connection<Provider>;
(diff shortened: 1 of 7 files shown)


tests/connection_pool.cpp
```diff
@@ -1,9 +1,13 @@
...
 GTEST("ozo::make_connection_pool") {
     SHOULD("not throw") {
-        EXPECT_NO_THROW(ozo::make_connection_pool("conn info string", 1, 1));
+        boost::asio::io_context io;
+        const ozo::connection_info<> conn_info(io, "conn info string");
+        const ozo::connection_pool_config config;
+        EXPECT_NO_THROW(ozo::make_connection_pool(conn_info, config));
     }
-}
+}
```
elsid: Left new line at the end of file.

include/ozo/impl/read.h
```diff
@@ -2,19 +2,20 @@
...
 using detail::istream;
 
 template <typename T>
 inline Require<Integral<T>, istream&> read(istream& in, T& out) {
-    T tmp;
-    in.read(reinterpret_cast<char*>(std::addressof(tmp)), sizeof out);
+    detail::typed_buffer<T> buf;
+    in.read(buf.raw, buf.size);
     if (!in) {
         throw system_error(error::unexpected_eof);
```
elsid: There is no test for this path.

include/ozo/connection_pool.h
```diff
@@ -32,35 +34,54 @@ struct pooled_connection {
...
+            using ::boost::asio::asio_handler_invoke;
+            asio_handler_invoke(std::forward<Func>(f), std::addressof(ctx->handler_));
+        }
+    };
+
     template <typename Handle>
     void operator ()(error_code ec, Handle&& handle) {
         if (ec) {
-            return handler_(std::move(ec), pooled_connection_ptr<OidMap, Statistics>{});
+            return handler_(std::move(ec), connection_ptr{});
```
elsid, line 84: We need tests for this case.

include/ozo/connection_pool.h
```diff
@@ -32,35 +34,54 @@ struct pooled_connection {
...
         }
 
-        auto conn = std::make_shared<pooled_connection<OidMap, Statistics>>(std::move(handle));
+        auto conn = std::make_shared<connection>(std::forward<Handle>(handle));
         if (!conn->empty() && connection_good(conn)) {
             return handler_(std::move(ec), std::move(conn));
```
elsid, line 89: This path doesn't have any tests.

elsid (changes requested):

include/ozo/connection_pool.h
thed636: **Done**

include/ozo/connection_pool.h
thed636: **Done**

include/ozo/impl/read.h
thed636: **Done**

include/ozo/connection_pool.h
```diff
@@ -8,59 +8,88 @@
...
     ~pooled_connection() {
-        if (connection_bad(*this)) {
+        if (!empty() && connection_bad(*this)) {
             handle_.waste();
         }
     }
 
     friend auto& unwrap_connection(pooled_connection& conn) noexcept {
-        return *(conn.handle_);
+        return unwrap_connection(*(conn.handle_));
```
elsid, line 43: You shouldn't do this function noexcept when any of called from there could throw exception, if you don't want to ge unexpected std::terminate. So noexcept specifier should depend on it. Something like this:
```c++
     friend auto& unwrap_connection(pooled_connection& conn) noexcept(noexcept(unwrap_connection(*(conn.handle_)))) {
        return unwrap_connection(*(conn.handle_));
     }
```

Probably we should check our code and fix it in other places. Or do we really need noexcept here?

include/ozo/connection_pool.h
```diff
@@ -70,68 +99,51 @@ struct pooled_connection_wrapper {
...
     "pooled_connection_ptr is not a Connectable concept");
 
-static_assert(ConnectionProvider<impl::pooled_connection_ptr<empty_oid_map, no_statistics>>,
+static_assert(ConnectionProvider<impl::pooled_connection_ptr<impl::connection<empty_oid_map, no_statistics>>>,
     "pooled_connection_ptr is not a ConnectionProvider concept");
 
-template <typename Connector, typename OidMap, typename Statistics>
+struct connection_pool_config {
+    std::size_t capacity = 1;
+    std::size_t queue_capacity = 1;
```
elsid: Maybe we should put here more reasonable defaults. Because this looks like very restricted case. You can have only one connection and have only one pending request. This will not work in any of ours usual cases. I think 10 and 100 will be more useful.

elsid (changes requested):

include/ozo/connection_pool.h
thed636: Yes, you right, we need to decide is 'noexcept' needed with unwrap_connection(). Now noexcept presents in connection.h for default unwrap_connection() implementation so that's why it presents here. 

let's fix it in separate PR, I'll make a task.

include/ozo/connection_pool.h
thed636: **Done**

elsid (dismissed):

elsid (approved):

PR #46, opened 2018-06-19
Stream io

thed636:
Binary serialization via std::ostream interface.

### DIFF
--- a/include/ozo/binary_query.h
+++ b/include/ozo/binary_query.h
@@ -19,25 +19,25 @@
 
 namespace ozo {
 
-template <class BufferAllocatorT, class OidMapImplT, class ... ParamsT>
+template <class TextT, class ParamsT, class OidMapT, class BufferAllocatorT>
 class binary_query {
 public:
-    static constexpr std::size_t params_count = sizeof ... (ParamsT);
+    static_assert(HanaTuple<ParamsT>, "Params must be hana::tuple");
+    static_assert(OidMap<OidMapT>, "OidMapT must be ozo::oid_map_t");
+    static_assert(QueryText<TextT>, "Text must be QueryText concept");
 
     using buffer_allocator_type = BufferAllocatorT;
-    using oid_map_type = oid_map_t<OidMapImplT>;
+    using oid_map_type = OidMapT;
+    using text_type = TextT;
+    using params_type = ParamsT;
 
-    binary_query(std::string_view text)
-        : impl(make_impl(buffer_allocator_type(), text, oid_map_type())) {}
+    static constexpr auto params_count = decltype(hana::length(std::declval<params_type>()))::value;
 
-    binary_query(std::string_view text, const oid_map_type& oid_map, const ParamsT& ... params)
-        : impl(make_impl(buffer_allocator_type(), text, oid_map, params ...)) {}
-
-    binary_query(const buffer_allocator_type& buffer_allocator, std::string_view text, const oid_map_type& oid_map, const ParamsT& ... params)
-        : impl(make_impl(buffer_allocator, text, oid_map, params ...)) {}
+    binary_query(text_type text, const params_type& params, const buffer_allocator_type& buffer_allocator, const oid_map_type& oid_map)
+        : impl(make_impl(std::move(text), params, oid_map, buffer_allocator)) {}
 
     constexpr const char* text() const noexcept {
-        return std::data(impl->text);
+        return to_const_char(impl->text);
     }
 
     constexpr const ::Oid* types() const noexcept {
@@ -62,15 +62,15 @@ private:
     using buffer_type = std::vector<char, buffer_allocator_type>;
 
     struct impl_type {
-        std::string_view text;
+        text_type text;
         buffer_type buffer;
         std::array<::Oid, params_count> types;
         std::array<int, params_count> formats;
         std::array<int, params_count> lengths;
         std::array<const char*, params_count> values;
 
-        impl_type(std::string_view text, const buffer_allocator_type& buffer_allocator)
-            : text(text), buffer(buffer_allocator) {}
+        impl_type(text_type text, const buffer_allocator_type& buffer_allocator)
+            : text(std::move(text)), buffer(buffer_allocator) {}
 
         impl_type(const impl_type&) = delete;
         impl_type(impl_type&&) = delete;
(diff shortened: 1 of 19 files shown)


include/ozo/impl/write.h
```diff
@@ -0,0 +1,72 @@
...
+#include <boost/hana/members.hpp>
+#include <boost/hana/tuple.hpp>
+
+namespace ozo::impl {
+
+using detail::ostream;
+
+template <typename T>
+inline Require<Integral<T> && sizeof(T) == 1, ostream&>
+write(ostream& out, T in) {
```
elsid, line 19: Use auto for return type, don't put function name on a new line:
```c++
inline auto write(ostream& out, T in)
    -> Require<Integral<T> && sizeof(T) == 1, ostream&> {
```

include/ozo/impl/write.h
```diff
@@ -0,0 +1,72 @@
...
+inline Require<FloatingPoint<T>, ostream&> write(ostream& out, T in) {
+    return write(out, detail::to_integral(in));
+}
+
+inline ostream& write(ostream& out, bool in) {
+    return write(out, (in ? char(1) : char(0)));
+}
+
+template <typename T>
+inline Require<FusionSequence<T>&&!HanaSequence<T>&&!HanaStruct<T>, ostream&>
```
elsid: Put spaces between operands and operators to make more readable.

include/ozo/detail/ostream.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <vector>
+#include <ostream>
+
+namespace ozo::detail {
+
+using std::ostream;
```
elsid: Why?

include/ozo/binary_query.h
```diff
@@ -67,13 +67,15 @@ class binary_query {
     struct impl_type {
         std::string_view text;
         buffer_type buffer;
+        ozo::detail::ostreambuf ostreambuf;
+        ozo::ostream stream;
```
elsid: Isn't it suppose to be defined somewhere else? This type should contain only what is required to make request using libpq, not to construct it. Streambuf and stream interface is not required to make request.

I can suggest divide field_proxy into field_meta_proxy and field_value_proxy and put stream into field_meta_proxy.
```c++
    template <std::size_t field>
    class field_meta_proxy {
    public:
        field_meta_proxy(impl_type& result, std::ostream& stream)
            : result(result), stream_(stream) {}

        constexpr void set_type(::Oid value) noexcept;
        constexpr void set_format(int value) noexcept;
        constexpr void set_length(int value) noexcept;
        constexpr auto& buffer() noexcept;
        constexpr auto& stream() noexcept {
            return stream_;
        }

    private:
        impl_type& result;
        std::ostream& stream_;
    };
    template <std::size_t field>
    class field_value_proxy {
    public:
        field_value_proxy(impl_type& result) : result(result) {}

        constexpr void set_value_by_offset(const std::size_t offset) noexcept {
            result.values[field] = std::data(result.buffer) + offset;
        }
        constexpr void set_null_value() noexcept {
            result.values[field] = nullptr;
        }

    private:
        impl_type& result;
    };
```
then make_impl should look like this:
```c++
static constexpr auto make_impl(...) {
        // ...
        std::streambuf buffer(result->buffer);
        std::ostream stream(&buffer);
        // ...
                field_meta_proxy<field> proxy(*result, stream);
        // ...
                    field_value_proxy<field>(*result).set_value_by_offset(offset);
                    // ...
                    field_value_proxy<field>(*result).set_null_value();
        // ...
}
```

elsid (changes requested):

include/ozo/impl/write.h
thed636: Well, to be honest, I don't remember any conventions about this. At least we have same style in "connection.h", "binary_deserealization.h" and so on. So if you have a suggestion to use this style of return type with Require statement emulation the best choice is to tell us why it is better and get a convention between all of us.

include/ozo/impl/write.h
thed636: **DONE**

include/ozo/detail/ostream.h
thed636: Why std::ostream or what? If you are asking about why alias (or using) is needed - the answer is - in the future, I think we will need our own stream or extensions ~~e.g., for strings on which we can make move semantic optimisation, since no byte order conversion is need.~~ (well, strings is not good example - this customisation point is in the different place).

include/ozo/binary_query.h
thed636: Thanks for the suggestion, I'll try to implement it my best!

include/ozo/detail/ostream.h
elsid: But why we can't add it in the future? Interface should be the same with std::ostream.

thed636: Sure, we can. I just did the same thing as for istream.h. The idea was to reserve place for custom stream implementation if it will be needed.

include/ozo/binary_serialization.h
```diff
@@ -2,107 +2,58 @@
...
+template <typename In, typename = std::void_t<>>
+struct send_impl{
+    template <typename M>
+    static ostream& apply(ostream& out, const oid_map_t<M>&, const In& in) {
+        return write(out, in);
+    }
+};
+
+template <class M, class In>
+inline std::ostream& send(std::ostream& out, const oid_map_t<M>& oid_map, const In& in) {
```
elsid: Replace std::ostream -> ostream.

include/ozo/detail/ostream.h
elsid: If you want to do this to restrict clients from use of std::ostream in their code, then better idea is to make adaptor type. Because now code with std::ostream is still correct for compiler, therefore when you will change this type later, clients could suffer.

include/ozo/binary_serialization.h
thed636: **Done**

include/ozo/detail/ostream.h
thed636: **Done**

elsid (approved):

include/ozo/impl/write.h
elsid: I don't have any good reason for my suggestion now.

thed636: For now I discovered the only one good reason - in case of something like this:

```c++
template <typename T1, typename T2>
inline auto foo(very_long<T1, and<sopthisticated_template_exprassion<T>>>& v) -> decltype(bar(v)) {
    return bar(v);
}
```

PR #48, opened 2018-07-10
Move tests from GUnit to GTest 

thed636:
Since we have problems to use GUnit on Windows and internally we need to move to GTest.
Here are commits for such movement.
There are also some refactoring commits which help to move to GTest.


tests/async_connect.cpp
```diff
@@ -29,175 +30,174 @@ using fixture = fixture_impl<empty_oid_map>;
 
 template <typename OidMap>
 auto make_fixture(OidMap) { return fixture_impl<OidMap>{}; }
+using namespace testing;
+using ozo::error_code;
 
-GTEST("ozo::async_connect()") {
-    using namespace testing;
-    using ozo::error_code;
+struct async_connect : ::testing::Test {};
```
elsid: Why do you explicitly use namespace `testing`?

elsid (changes requested):

include/ozo/impl/read.h
```diff
@@ -24,6 +24,9 @@ inline Require<RawDataWritable<T>, istream&> read(istream& in, T& out) {
 template <typename T>
 inline Require<Integral<T> && sizeof(T) == 1, istream&> read(istream& in, T& out) {
     out = in.get();
+    if (!in) {
+        throw system_error(error::unexpected_eof);
```
elsid, line 28: There is [exceptions](https://en.cppreference.com/w/cpp/io/basic_ios/exceptions) method for such stuff.

tests/async_get_result.cpp
```diff
@@ -35,399 +36,414 @@ using ozo::impl::query_state;
 using namespace testing;
 using ozo::error_code;
 
-GTEST("ozo::async_get_result_op::perform()") {
+struct async_get_result_op : testing::Test {
```
elsid: Why do you both use `using namespace testing` and explicit name for this namespace?

elsid (changes requested):

dpotepalov (dismissed):

include/ozo/impl/read.h
thed636: Well, in case of using ```exception()``` method, ```std::ios_base::failure```  will be thrown which ~~is not compatible with ```error_code``` and~~ (oops, my bad, valid until C++11 only) IMO completely meaningless - no useful info can be added at all. In case of throwing ```system_error``` here, additional string with information can be added in the future as soon as it needed.

PS: BTW ```std::ios_base::failure``` inherits for ```std::system_error``` and it differs from Boost variant which we are using doe to be compatible with Boost.Asio.

tests/async_get_result.cpp
thed636: Just a habit. Any problem with it? Should I remove it?

elsid: The problem is one will have one more question "Why? Is there any hidden purpose?". Remove one of these.

thed636: Ok, **done**

elsid (approved):

PR #56, opened 2018-07-27
Connection refactoring

thed636:
Slightly refactored connection related concepts.

Remove Connectable concept as unused. A wrapped connection is a connection too.
Move customization some points to metafunction implementations (as result - reduced amount of code)
Default async_get_connection() implementation now uses provider's async_get_connection() member function instead of operator().

So next step is to make pooled_connection as a connection but not a wrapper to avoid store io_context bound socket.

### DIFF
--- a/include/ozo/connection.h
+++ b/include/ozo/connection.h
@@ -304,24 +304,6 @@ inline decltype(auto) get_statistics(T&& conn) noexcept {
     return get_connection_statistics(unwrap_connection(std::forward<T>(conn)));
 }
 
-// Oid Map helpers
-
-template <typename T, typename C, typename = Require<Connectable<C>>>
-inline decltype(auto) type_oid(C&& conn) noexcept {
-    return type_oid<std::decay_t<T>>(
-        get_oid_map(std::forward<C>(conn)));
-}
-
-template <typename T, typename C, typename = Require<Connectable<C>>>
-inline decltype(auto) type_oid(C&& conn, const T&) noexcept {
-    return type_oid<std::decay_t<T>>(std::forward<C>(conn));
-}
-
-template <typename T, typename C, typename = Require<Connectable<C>>>
-inline void set_type_oid(C&& conn, oid_t oid) noexcept {
-    set_type_oid<T>(get_oid_map(std::forward<C>(conn)), oid);
-}
-
 
 template <typename ConnectionProvider, typename Enable = void>
 struct get_connectable_type {
--- a/include/ozo/connection.h
+++ b/include/ozo/connection.h
@@ -10,8 +10,6 @@
 
 namespace ozo {
 
-using impl::pg_conn_handle;
-
 using no_statistics = decltype(hana::make_map());
 
 /**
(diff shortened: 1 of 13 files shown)


include/ozo/connection.h
```diff
@@ -20,6 +20,62 @@ using no_statistics = decltype(hana::make_map());
...
+struct get_connection_oid_map_impl {
+    template <typename Conn>
+    static auto apply(Conn&& c) -> decltype(c.oid_map()) { return c.oid_map();}
+    template <typename Conn>
+    static auto apply(Conn&& c) -> decltype((c.oid_map_)) { return c.oid_map_;}
+};
+
+template <typename T>
+inline auto get_connection_oid_map(T&& conn) ->
+    decltype (get_connection_oid_map_impl<std::decay_t<T>>::apply(std::forward<T>(conn))){
```
elsid: Put `->` on this line and add indent. Now it is hard to distinguish this line from function body.

include/ozo/connection.h
```diff
@@ -20,6 +20,62 @@ using no_statistics = decltype(hana::make_map());
 template <typename, typename = std::void_t<>>
 struct is_connection : std::false_type {};
 
+template <typename T, typename = std::void_t<>>
+struct get_connection_oid_map_impl {
+    template <typename Conn>
+    static auto apply(Conn&& c) -> decltype(c.oid_map()) { return c.oid_map();}
+    template <typename Conn>
+    static auto apply(Conn&& c) -> decltype((c.oid_map_)) { return c.oid_map_;}
```
elsid: Why do you need additional parentheses?

include/ozo/connection.h
```diff
@@ -20,6 +20,62 @@ using no_statistics = decltype(hana::make_map());
...
+    decltype (get_connection_oid_map_impl<std::decay_t<T>>::apply(std::forward<T>(conn))){
+    return get_connection_oid_map_impl<std::decay_t<T>>::apply(std::forward<T>(conn));
+}
+
+template <typename T, typename = std::void_t<>>
+struct get_connection_socket_impl {
+    template <typename Conn>
+    static auto apply(Conn&& c) -> decltype(c.socket()) { return c.socket();}
+    template <typename Conn>
+    static auto apply(Conn&& c) -> decltype((c.socket_)) { return c.socket_;}
```
elsid: Why did you removed reference from return type?

elsid (changes requested):

include/ozo/native_conn_handle.h
```diff
@@ -0,0 +1,27 @@
...
+ * 2. By "std::default_delete" template specialization
+ *
+ * In the first case we lose the default constructor for the pointer.
+ * That's why the second way was chosen.
+ */
+namespace std {
+
+template <>
+class default_delete<PGconn> {
+public:
```
elsid: Use struct

elsid (changes requested):

include/ozo/connection.h
thed636: **Done**

include/ozo/native_conn_handle.h
thed636: **Done**

include/ozo/connection.h
thed636: To return by reference, not by value - it works like decltype(auto). There is no other way to return by reference using decltype in SFINAE manner.

include/ozo/connection.h
thed636: See the comment on `get_connection_oid_map_impl::apply()`

elsid (approved):

PR #57, opened 2018-08-09
Connection provider v2

elsid:
Alternative implementation to #55

### DIFF
--- a/include/ozo/connection.h
+++ b/include/ozo/connection.h
@@ -319,6 +319,20 @@ struct async_get_connection_impl {
     }
 };
 
+template <typename, typename = std::void_t<>>
+struct is_connection_source : std::false_type {};
+
+template <typename T>
+struct is_connection_source<T, std::void_t<decltype(
+    std::declval<T&>()(
+        std::declval<io_context&>(),
+        std::declval<std::function<void(error_code, connection_type<T>)>>()
+    )
+)>> : std::true_type {};
+
+template <typename T>
+constexpr auto ConnectionSource = is_connection_source<std::decay_t<T>>::value;
+
 template <typename T, typename Handler>
 inline auto async_get_connection(T&& provider, Handler&& handler)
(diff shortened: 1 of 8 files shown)


include/ozo/connection_info.h
```diff
@@ -11,29 +12,28 @@ template <
...
-    Statistics statistics_;
-
-    using connection = impl::connection_impl<OidMap, Statistics>;
+    std::string conn_str;
+    Statistics statistics;
 
 public:
+    using connection = impl::connection_impl<OidMap, Statistics>;
     using connection_type = std::shared_ptr<connection>;
+    using source_type = connection_info;
```
thed636: IMO **source_type** is suitable only for the **connector** class because in other cases this type points to different entities what is misleading. From this PR I can see no application for this type outside of **connector** class. If you have any ideas about the future of this type, please describe it in few words.

include/ozo/connector.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
+
+#include <ozo/asio.h>
+#include <ozo/connection.h>
+
+namespace ozo {
+
+template <typename Base, typename ... Args>
+class connector {
```
thed636, line 9: I think the **ConnectionSource** concept check is needed for the **Base** like
```C++
static_assert(ConnectionSource<Base>, "is not a ConnectionSource");
```

include/ozo/connection.h
```diff
@@ -319,6 +319,16 @@ struct async_get_connection_impl {
     }
 };
 
+template <typename, typename = std::void_t<>>
+struct is_connection_source : std::false_type {};
+template <typename T>
+struct is_connection_source<T, std::void_t<
```
thed636: IMO it is better to make stronger concept check like for **ConnectionProvider**, something like this
```C++
template <typename T>
struct is_connection_source<T, std::void_t<decltype(
    std::declval<T&>()(
        std::declval<io_context&>(), 
        std::declval<std::function<void(error_code, connection_type<T>)>>()
    )
)>> : std::true_type {};
```

thed636 (changes requested):

thed636 (approved):

PR #58, opened 2018-08-10
Transaction

elsid:


include/ozo/transaction.h
```diff
@@ -0,0 +1,90 @@
...
+    using namespace ozo::literals;
+    using signature_t = void (error_code, impl::transaction<connection_type<P>>);
+
+    const auto result = std::make_shared<ozo::result>();
+    async_completion<CompletionToken, signature_t> init(token);
+
+    request(std::forward<P>(provider), "BEGIN"_SQL, *result,
+        [handler = init.completion_handler, result] (error_code ec, auto&& conn) mutable {
+            handler(std::move(ec), impl::make_transaction(std::move(conn)));
+        });
```
thed636: Lambda leads to the handler context loss. So you must use a wrapper with overloaded asio_handler_invoke() to preserve the handler context. It is sad, but - no other way.

include/ozo/transaction.h
```diff
@@ -0,0 +1,90 @@
...
+    T conn_;
+
+    friend auto& unwrap_connection(transaction& self) noexcept {
+        return unwrap_connection(self.conn_);
+    }
+
+    friend const auto& unwrap_connection(const transaction& self) noexcept {
+        return unwrap_connection(self.conn_);
+    }
+};
```
thed636: We need to do something at the transaction object destruction, in case the transaction is still in progress, to avoid such things as returning connection with not finished transaction into a connection pool, for example.

include/ozo/transaction.h
```diff
@@ -0,0 +1,90 @@
...
+
+template <typename Conn, typename = Require<Connection<Conn>>>
+auto make_transaction(Conn&& conn) {
+    return transaction<Conn> {std::forward<Conn>(conn)};
+}
+
+} // namespace impl
+
+template <typename P, typename CompletionToken, typename = Require<ConnectionProvider<P>>>
+auto transaction(P&& provider, CompletionToken&& token) {
```
thed636: IMO better name for the function is 'begin()' - like an SQL command.

thed636 (changes requested):

include/ozo/transaction.h
elsid: Fixed

include/ozo/transaction.h
elsid: Fixed

include/ozo/transaction.h
elsid: Fixed

include/ozo/impl/transaction.h
```diff
@@ -0,0 +1,52 @@
...
+    ~transaction() {
+        const auto c = std::move(counter);
+        if (c.use_count() == 1) {
+            close_connection(std::move(connection));
+        }
+    }
+
+    void take_connection(T& out) {
+        out = std::move(connection);
+        counter.reset();
```
thed636: Unfortunately, this will not work, you just decrement shared_ptr counter in this line, no more.  A copy of the transaction object still has a connection and counter decreased by one.

include/ozo/impl/transaction.h
```diff
@@ -0,0 +1,52 @@
...
+              counter(std::make_shared<std::monostate>()) {}
+
+    ~transaction() {
+        const auto c = std::move(counter);
+        if (c.use_count() == 1) {
+            close_connection(std::move(connection));
+        }
+    }
+
+    void take_connection(T& out) {
```
thed636: It seems like such method must be private and `end_transaction()` must be the class method, or we need some extra logic to do not allow a user to extract connection manually.

<details><summary><b>Something boring like this</b></summary><p>

```c++
    auto reset(error_code& ec) {
        if (transaction in progress) {
            ec = error::transaction_in_progress;
            return nullptr;
       }
        //...
       return std::move(connection);
    }

    auto reset() {
        error_code ec;
        auto out = reset(ec);
        if (ec) {
            throw system_error(ec);
        }
        return out;
    }
```

</p></details>

include/ozo/impl/transaction.h
```diff
@@ -0,0 +1,52 @@
...
+        return unwrap_connection(self.connection);
+    }
+
+    friend const auto& unwrap_connection(const transaction& self) noexcept {
+        return unwrap_connection(self.connection);
+    }
+
+private:
+    T connection;
+    std::shared_ptr<std::monostate> counter;
```
thed636: The only solution is to have a shared pointer to the connection object with a custom destructor, something like 

```c++
struct context_dtor {
    void opeartor()(T* conn) {         
        if (conn) {
            close_connection(*conn);
        }
    }
};

std::shared_ptr<T> ctx;

//...

transaction(T connection) : ctx{new T(connection), context_dtor{}}
```

include/ozo/transaction.h
```diff
@@ -0,0 +1,68 @@
...
+    using namespace ozo::literals;
+    return start_transaction(
+        std::forward<T>(provider),
+        "BEGIN"_SQL,
+        std::forward<CompletionToken>(token)
+    );
+}
+
+template <typename T, typename Query, typename CompletionToken, typename = Require<Connection<T>>>
+auto end_transaction(impl::transaction<T>&& transaction, Query&& query, CompletionToken&& token) {
```
thed636: `Require<Connection<T>>` requirement is not needed because transaction requires Connection for underlying type.

include/ozo/transaction.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
+
+#include <ozo/impl/async_start_transaction.h>
+#include <ozo/impl/async_end_transaction.h>
+
+namespace ozo {
+
+template <typename T, typename Query, typename CompletionToken, typename = Require<ConnectionProvider<T>>>
+auto start_transaction(T&& provider, Query&& query, CompletionToken&& token) {
```
thed636: IMO it is not a good idea allow to construct transaction object without an actual transaction. I mean it is not a part of public interface.

include/ozo/binary_deserialization.h
```diff
@@ -232,4 +232,23 @@ basic_result<T>& recv_result(basic_result<T>& in, const oid_map_t<M>&, basic_res
     return out;
 }
 
+template <typename T, typename M>
+basic_result<T>& recv_result(basic_result<T>& in, const oid_map_t<M>& oid_map, std::reference_wrapper<basic_result<T>> out) {
+    return recv_result(in, oid_map, out.get());
+}
+
+template <typename M>
+basic_result<shared_native_result_handle>& recv_result(basic_result<native_result_handle>& in,
```
thed636: Do you really see an application for the shared result? I mean, we provide the result as-is only if a user wants to do something non-standard with raw things. It's ok, but is it really needed to provide more functionality in this direction?

include/ozo/binary_deserialization.h
```diff
@@ -232,4 +232,23 @@ basic_result<T>& recv_result(basic_result<T>& in, const oid_map_t<M>&, basic_res
     return out;
 }
 
+template <typename T, typename M>
+basic_result<T>& recv_result(basic_result<T>& in, const oid_map_t<M>& oid_map, std::reference_wrapper<basic_result<T>> out) {
```
thed636, line 223: This is strange, we already use basic_result as an output argument which being accepted by non-constant reference only, what is the point to support a reference wrapper for reference argument?

include/ozo/binary_deserialization.h
```diff
@@ -232,4 +232,23 @@ basic_result<T>& recv_result(basic_result<T>& in, const oid_map_t<M>&, basic_res
...
+basic_result<shared_native_result_handle>& recv_result(basic_result<native_result_handle>& in,
+        const oid_map_t<M>&, basic_result<shared_native_result_handle>& out) {
+    out = basic_result<shared_native_result_handle>(
+        shared_native_result_handle(in.handle().release(), std::default_delete<PGresult> {})
+    );
+    return out;
+}
+
+template <typename T, typename M>
+no_result& recv_result(basic_result<T>&, const oid_map_t<M>&, no_result& out) {
```
thed636: The best choice is to provide execute() method - in this case, no extra recv_result() is needed.

<details><summary><b>Something like this based on request</b></summary><p>

```c++
template <typename Query, typename Out, typename Handler>
inline auto make_async_execute_op(Query&& query, Handler&& handler) {
    auto do_nothing = [](auto, auto){};
    return impl::async_request_op<decltype(do_nothing),
            std::decay_t<Query>, std::decay_t<Handler>>{
        do_nothing,
        std::forward<Query>(query),
        std::forward<Handler>(handler)
    };
}

template <typename P, typename Q, typename Out, typename Handler>
inline void async_execute(P&& provider, Q&& query, Handler&& handler) {
    static_assert(ConnectionProvider<P>, "is not a ConnectionProvider");
    static_assert(Query<Q> || QueryBuilder<Q>, "is neither Query nor QueryBuilder");
    async_get_connection(std::forward<P>(provider),
        impl:: make_async_execute_op(
            std::forward<Q>(query),
            std::forward<Handler>(handler)
        )
    );
}
```

</p></details>

thed636 (changes requested):

include/ozo/transaction.h
elsid: There are different ways how to start transaction in PostgreSQL.

https://www.postgresql.org/docs/current/static/sql-begin.html
https://www.postgresql.org/docs/current/static/sql-start-transaction.html

Since we support only begin now, why do we restrict users to use another one?

include/ozo/binary_deserialization.h
elsid: If you need create handler with only result, you need to make `shared_ptr` wrapper over `ozo::result`. `ozo::shared_result` allow not to do it, and use only one pointer.

include/ozo/binary_deserialization.h
elsid: With this commit you can't anymore. `async_request_out_handler` shouldn't store raw reference. For example if you pass `std::shared_ptr` to `make_async_request_op` before this commit, you will have only reference to it, not a copy. This example shows the problem:
```c++
ozo::basic_result<std::shared_ptr<PGresult>> res(
    std::shared_ptr<PGresult>(reinterpret_cast<PGresult*>(1), [] (auto& v) { v = 0; })
);

EXPECT_EQ(res.handle().get(), reinterpret_cast<PGresult*>(1));

auto op = ozo::impl::make_async_request_op("SELECT 1"_SQL + " + 1"_SQL, res, [] (auto ...) {});

res.handle().reset();

EXPECT_EQ(op.out_.out.handle().get(), nullptr);
// EXPECT_EQ(op.out_.out.handle().get(), reinterpret_cast<PGresult*>(1)); // desired behaviour
```

My opinion is to force users to specify whether they want to copy result or to pass reference.

include/ozo/impl/transaction.h
elsid: This type is a part of implementation. Users should use its methods directly only if they really know what they want. We have a lot of types in this project with public fields and it is ok, why this type is different?

include/ozo/impl/transaction.h
elsid: Ok, but I can do it another way. `transaction` destructor can do same logic without overhead for deleter std::function. Also if connection is a `shared_ptr` we don't need any extra one.

include/ozo/transaction.h
elsid: Fixed

include/ozo/transaction.h
thed636: As it is pointed in the second link `This is the same as the BEGIN command.`. So it seems like there is no restriction for the user if we use only `begin` as transaction start keyword. The `transaction_mode` options can be arguments of the transaction start function.

There is another thing:
> In the standard, it is not necessary to issue START TRANSACTION to start a transaction block: any SQL command implicitly begins a block. PostgreSQL's behavior can be seen as implicitly issuing a COMMIT after each command that does not follow START TRANSACTION (or BEGIN), and it is therefore often called “autocommit”. Other relational database systems might offer an autocommit feature as a convenience.

IMO it is better explicitly use `begin` keyword because it works the same way with both autocommit on and off modes.

include/ozo/binary_deserialization.h
thed636: But what for? It must be a non-const reference (or rvalue reference if we construct inserter right on the function call) always because the user is responsible for the output object's lifetime and condition. In your example it is going something strange - the user is trying to destroy the output object on purpose. Why do you want something like a copy of std::shared_ptr instead of reference. I mean what is the aim of this change? It is not clear from the example. Are there more reasonable things like dangling references inside the library or something like this, but without user's manipulations with the output object?

> My opinion is to force users to specify whether they want to copy result or to pass reference.

User do not want to copy result, because result is the output and the only way to get output is to have original result object or reference to. There is just no option at all.

elsid: > the user is trying to destroy the output object on purpose

If you allocate `res` on a stack and don't copy anywhere else same thing is happen.

If I do not capture res into lambda, I have SIGSEGV. It is not obvious not to copy shared_ptr, but take only a reference.
```c++
io.post([&] {
    ozo::shared_result res;
    ozo::request(ozo::make_connector(conn_info, io), "SELECT 1"_SQL, res,
        [] (auto, auto) {});
});
```

Another example. I pass iterator using lvalue, and surprise is request takes only reference to it. So I have SIGSEGV.
```c++
rows_of<std::vector<std::string>> res(1);
io.post([&] {
    auto it = res.begin();
    ozo::request(ozo::make_connector(conn_info, io), "SELECT "_SQL + foos, it,
            [](auto, auto) {});
});
```

include/ozo/binary_deserialization.h
elsid: Fixed

include/ozo/binary_deserialization.h
elsid: Removed

include/ozo/transaction.h
elsid: Done

include/ozo/binary_deserialization.h
thed636: Well, this is the behavior we define via the interface. A user has to control the lifetime of the output object. That's it. Via such behavior, we do not obligate user to use shared pointers or something like this in case if it is not needed(e.g. Boost.Coroutine or `std::future` or just to make concurrent requests in synchronous code and do not start a different thread for `io.run()`).  If a user decides to use a callback-style asynchronous code, then he or she has to control objects lifetime, as usual. This is common practice.

tests/request_integration.cpp
```diff
@@ -32,7 +32,7 @@ TEST(request, should_return_error_and_bad_connect_for_invalid_connection_info) {
     ozo::connection_info<> conn_info("invalid connection info");
 
     ozo::result res;
-    ozo::request(ozo::make_connector(conn_info, io), "SELECT 1"_SQL + " + 1"_SQL, res,
+    ozo::request(ozo::make_connector(conn_info, io), "SELECT 1"_SQL + " + 1"_SQL, std::ref(res),
```
thed636, line 35: As you can see you need to point std::ref everywhere because of no other options with `ozo::result`.

tests/impl/async_end_transaction.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    auto transaction = ozo::impl::transaction<decltype(conn)>(std::move(conn));
+
+    Sequence s;
+
+    EXPECT_CALL(executor, dispatch(_)).InSequence(s).WillOnce(InvokeArgument<0>());
+    EXPECT_CALL(strand_service, get_executor()).InSequence(s).WillOnce(ReturnRef(strand));
+
+    // Send query params
+    EXPECT_CALL(connection, set_nonblocking()).InSequence(s).WillOnce(Return(0));
```
thed636: This is very low-level details which are not the test subject. Here should be mocked async_execute() call and no more.

tests/impl/async_get_result.cpp
```diff
@@ -161,28 +181,30 @@ TEST_F(async_get_result, should_post_callback_with_error_if_consume_input_failed
...
 
     // Consume input
     EXPECT_CALL(m.connection, consume_input()).InSequence(s).WillOnce(Return(0));
 
     // Cancel all io
-    EXPECT_CALL(m.socket, cancel(_)).WillOnce(Return());
+    EXPECT_CALL(m.socket, cancel(_)).InSequence(s).WillOnce(Return());
 
-    // Post callback with error
+    // Post callback with
```
thed636: ...error? Or something else?

tests/impl/async_get_result.cpp
```diff
@@ -161,28 +181,30 @@ TEST_F(async_get_result, should_post_callback_with_error_if_consume_input_failed
...
+    EXPECT_CALL(m.callback, get_executor()).WillOnce(Return(ozo::tests::executor {&m.callback_executor}));
+    EXPECT_CALL(m.callback_executor, on_work_started()).InSequence(s).WillOnce(Return());
     EXPECT_CALL(m.executor, post(_)).InSequence(s).WillOnce(InvokeArgument<0>());
-    EXPECT_CALL(m.strand, dispatch(_)).InSequence(s).WillOnce(InvokeArgument<0>());
-    EXPECT_CALL(m.callback, context_preserved()).InSequence(s).WillOnce(Return());
+    EXPECT_CALL(m.callback_executor, dispatch(_)).InSequence(s).WillOnce(InvokeArgument<0>());
     EXPECT_CALL(m.callback, call(error_code{ozo::error::pg_consume_input_failed}, _))
         .InSequence(s).WillOnce(Return());
+    EXPECT_CALL(m.callback_executor, on_work_finished()).InSequence(s).WillOnce(Return());
+    EXPECT_CALL(m.strand, on_work_finished()).InSequence(s).WillOnce(Return());
```
thed636: on_work_started()/on_work_finished() are really important for the tests? You can add for them default behaviour in fixture to do not saturate tests with not important things.

tests/impl/async_start_transaction.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+TEST_F(async_start_transaction, should_call_async_execute) {
+    *conn->handle_ = native_handle::good;
+
+    Sequence s;
+
+    EXPECT_CALL(executor, dispatch(_)).InSequence(s).WillOnce(InvokeArgument<0>());
+    EXPECT_CALL(strand_service, get_executor()).InSequence(s).WillOnce(ReturnRef(strand));
+
+    // Send query params
```
thed636: Too low level, same as here https://github.com/yandex/ozo/pull/58/files/8c25fe0a548b057b7092cc38d9e9eae16c0b2bbb#diff-b2000bf70b252ad7f99cd703250f1f8dR39

In this tests async_execute() call should be mocked, that is why you need different connection mocks.

thed636 (changes requested):

tests/impl/async_start_transaction.cpp
elsid: Done

tests/impl/async_get_result.cpp
elsid: Done

thed636 (dismissed):

thed636 (approved):

PR #61, opened 2018-08-16
Less code, more docs

thed636:
Remove some code and add the very first version of How To.


README.md
```diff
@@ -10,7 +10,7 @@ The library leverages the power of template metaprogramming, providing convenien
...
+Since the project is on early state of developmet it lacks of documentation. We understand the importance of good docs and are working hard on this problem. Complete documentation is on the way, but now: look at our [HOW TO](docs/howto.md), then [unit tests](tests/request_integration.cpp) which can help you understand the main use-cases of OZO. We also have a C++Now'18 talk about OZO. The persentation is [here](https://github.com/boostcon/cppnow_presentations_2018/blob/master/05-09-2018_wednesday/design_and_implementation_of_dbms_asynchronous_client_library__roman_siromakha__cppnow_05092018.pdf). Video will be provided later at [C++Now 2018 playlist](https://www.youtube.com/watch?v=QFIOE1jKv30&list=PL_AKIMJc4roVSbTTfHReQTl1dc9ms0lWH).
```
elsid: Let's split this paragraph by lines of ~100 symbols.

elsid: `developmet` -> `development`
`Video will be provided later at` it is already there.

docs/howto.md
```diff
@@ -0,0 +1,174 @@
...
+
+    // Rows which accepts integer and nullable string columns in the sequence
+    // It is an alias on std::vector of std::tuple
+    ozo::rows_of<std::int64_t, std::optional<std::string>> rows;
+
+    // Connection info with host and port to coonect to
+    ozo::connection_info<> conn_info("host=... port=...");
+
+    // Our query statement
+    const auto query = "SELECT id, name FROM users_info WHERE amount>="_SQL + std::int64_t(25);
```
elsid: Add `using namespace ozo::literals`.

docs/howto.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+Here we define result type as we want to see it. Practically `ozo::rows_of` is an alias on `std::vector<std::tuple<...>>`. And `ozo::into` is an alias on `std::back_inserter`. So `request()` function will fill this vector of tuples according to the database response rows. It is very important to preserve the same order of fields in request and types in the tuple (it is a little bit annoyng, but there is a way to avoid it via [Boost.Hana](https://www.boost.org/doc/libs/1_66_0/libs/hana/doc/html/index.html#tutorial-introspection-adapting) or [Boost.Fusion](https://www.boost.org/doc/libs/1_66_0/libs/fusion/doc/html/fusion/adapted.html) structure adaptation). We can see `std::optional` at second position in the tuple. The reason is what `name` field can be _NULL_. So the empty optional is the _NULL_ representation (see nullables for more info about it). If you ommit `std::optional` then in case of _NULL_ there will be a result deserialization error. The first argument can not be a _NULL_, so here we do not to bother about it. There is no mistake to expect nullable type for non-nullable result, but the opposite leads to run-time error.
```
elsid: Split this paragraph by lines of ~100 symbols.

elsid: `annoyng` -> `annoying`

docs/howto.md
```diff
@@ -0,0 +1,174 @@
...
+
+It is a good question! Since we are using binary protocol and have serialization/deserialization system we also have a type system. It is easy and extandable. For build-in types you can look at [ozo/type_traits.h](../include/ozo/type_traits.h) for difinitions like:
+
+```c++
+OZO_PG_DEFINE_TYPE(std::vector<char>, "bytea", BYTEAOID, dynamic_size)
+
+OZO_PG_DEFINE_TYPE_AND_ARRAY(boost::uuids::uuid, "uuid", UUIDOID, 2951, bytes<16>)
+```
+
+`OZO_PG_DEFINE_TYPE` defines a C++ to built-in PostgreSQL type mapping. First argument here is a C++ type, second - PostgreSQL type name. Fird is type OID in PostgreSQL, fourth - type size which can be static for N bytes `bytes(N)`, or dynamic `dynamic_size`, like in this case.
```
elsid: `Fird` -> `First`

elsid (changes requested):

docs/howto.md
thed636: **Done**

docs/howto.md
thed636: **Done**

docs/howto.md
thed636: **Done**

README.md
thed636: **Done**

include/ozo/connection.h
```diff
@@ -12,20 +12,13 @@ namespace ozo {
...
-constexpr auto ConnectionWrapper = is_connection_wrapper<std::decay_t<T>>::value;
-
 /**
 * Overloaded version of unwrap function for Connection.
 * Returns connection object itself.
 */
 template <typename T>
 inline decltype(auto) unwrap_connection(T&& conn,
-        Require<!ConnectionWrapper<T>>* = 0) noexcept {
+        Require<!Nullable<T>>* = 0) noexcept {
```
elsid: `std::reference_wrapper` is not nullable but `unwrap_connection` is required to get value from it.

include/ozo/connection.h
```diff
@@ -63,6 +134,32 @@ struct get_connection_socket_impl {
     }
 };
 
+/**
+ * @ingroup Connection_Concept
+ * @brief Get the connection socket object
+ *
+ * Connection socket object getter. Connection must provide socket for IO. In general it must return
+ * reference to an instance of [boost::asio::posix::stream_descriptor](https://www.boost.org/doc/libs/1_67_0/doc/html/boost_asio/reference/posix__stream_descriptor.html).
```
elsid: We support boost 1.66 in CI, better use links to this version.

include/ozo/connection.h
```diff
@@ -91,6 +214,34 @@ struct get_connection_error_context_impl {
...
 };
 
+/**
+ * @ingroup Connection_Concept
+ * @brief Get the connection error context object
+ *
+ * Connection OZO error context getter. In many cases the `error_code` is a good thing, but sometimes
+ * you need an additional context which unhopefully can not be placed inside an `error_code` object. To
+ * solve this problem this additional error context is introduced. Usually it returns a reference to an
+ * std::string object. The error context is reseted inside get_connection implementation for Connection.
```
elsid: reseted -> reset

include/ozo/connection.h
```diff
@@ -344,20 +680,86 @@ struct is_connection_provider<T, std::void_t<decltype(
...
+    void async_get_connection(Handler&& h);
+ * @endcode
+ *
+ * where Handler must be a functor with such or compatible signature:
+ *
+ * @code
+    void (error_code ec, connection_type<ConnectionProvider> conn);
+ * @endcode
+ *
+ * This behavour can be customized via async_get_connection_impl specialization. E.g. for Connection
```
elsid: behavour -> behaviour
Use spelling checker, Luke!

elsid (changes requested):

include/ozo/connection.h
thed636: Well, this version of `unwrap_connection` does not work with `std::reference_wrapper` so as it didn't work in the previous version too. So nothing has changed. To support `std::reference_wrapper` the overloaded version of `unwrap_connection` is needed. Something like this (according to one of the next commits):
```c++
template <typename T>
struct unwrap_connection_impl<std::reference_wrapper<T>> {
    static constexpr decltype(auto) apply(std::reference_wrapper<T> ref) noexcept {
        return unwrap_connection(ref.get());
    }
};
```
It has to be done thoroughly but in a different pull request.

include/ozo/connection.h
thed636: **Done**

include/ozo/connection.h
thed636: **Done**

include/ozo/connection.h
thed636: **Done**

Thanks, I will.

elsid (approved):

PR #62, opened 2018-08-21
Request timeout

elsid:


include/ozo/impl/async_connect.h
```diff
@@ -135,17 +112,36 @@ struct async_connect_op {
         done(error::pq_connect_poll_failed);
     }
 
+    void done(error_code ec = error_code {}) {
+        decltype(auto) io = get_io_context(get_connection(context));
+
+        asio::post(io,
```
thed636: In this case you can post directly to executor without binding

```c++
asio::post(get_executor(), detail::bind(
    std::move(get_handler(context)), std::move(ec), std::move(get_connection(context)));
```

include/ozo/impl/async_connect.h
```diff
@@ -58,55 +61,29 @@ auto& get_executor(const connect_operation_context_ptr<Ts ...>& context) {
...
-
-        decltype(auto) io = get_io_context(get_connection(context));
-
-        asio::post(io, detail::bind(std::move(get_handler(context)), std::move(ec), std::move(get_connection(context))));
-    }
-
     void perform(const std::string& conninfo, const time_traits::duration& timeout) {
         if (error_code ec = start_connection(get_connection(context), conninfo)) {
-            return done(ec);
+            return cancel(ec);
```
thed636: What is the difference between `cancel` and `done` except executor? Is this separation really needed?

include/ozo/impl/async_request.h
```diff
@@ -80,12 +85,22 @@ inline void done(const request_operation_context_ptr<Ts...>& ctx, error_code ec)
     decltype(auto) conn = get_connection(ctx);
     error_code _;
     get_socket(conn).cancel(_);
-    post(get_connection(ctx), detail::bind(get_handler(ctx), std::move(ec), conn));
+    post(get_connection(ctx),
```
thed636: You can post it via the executer directly.

include/ozo/impl/async_request.h
```diff
@@ -80,12 +85,22 @@ inline void done(const request_operation_context_ptr<Ts...>& ctx, error_code ec)
...
+            get_executor(ctx),
+            detail::bind(get_handler(ctx), std::move(ec), conn)
+        )
+    );
 }
 
 template <typename ...Ts>
 inline void done(const request_operation_context_ptr<Ts...>& ctx) {
-    post(get_connection(ctx), detail::bind(get_handler(ctx), error_code{}, get_connection(ctx)));
+    post(get_connection(ctx),
```
thed636: You can post it via the executer directly.

include/ozo/request.h
```diff
@@ -5,14 +5,25 @@
 namespace ozo {
 
 template <typename P, typename Q, typename Out, typename CompletionToken, typename = Require<ConnectionProvider<P>>>
-inline auto request(P&& provider, Q&& query, Out&& out, CompletionToken&& token) {
+auto request(P&& provider, Q&& query, const time_traits::duration& timeout, Out&& out, CompletionToken&& token) {
```
thed636: Looks like Out must be a value in public interface, since we pass output only by value.

include/ozo/impl/async_request.h
```diff
@@ -315,7 +340,17 @@ struct async_request_op {
             return handler_(ec, std::move(conn));
         }
 
-        auto ctx = make_request_operation_context(std::move(conn), std::move(handler_));
+        ozo::strand<decltype(get_io_context(conn))> strand(get_io_context(conn));
+        auto ctx = make_request_operation_context(
+            std::move(conn),
+            detail::make_cancel_timer_handler(
```
thed636: `cancel_timer_handler` must be bound to the same stand as the `timeout_handler`. Now it seems like these handlers are bound to different strands. The first one - for the stand from this `ctx`, and the second one - for the strand from `operator()`.

include/ozo/impl/async_request.h
```diff
@@ -315,7 +340,17 @@ struct async_request_op {
...
+                detail::make_post_handler(std::move(handler_))
+            )
+        );
+
+        get_timer(get_connection(ctx)).expires_after(timeout_);
+        get_timer(get_connection(ctx)).async_wait(asio::bind_executor(strand,
```
thed636: According to the comment from above, it looks like here must be an executor from `ctx`:

```c++
get_timer(get_connection(ctx)).async_wait(asio::bind_executor(get_executor(ctx)
    detail::make_timeout_handler(get_socket(get_connection(ctx)))));
```

include/ozo/connection_pool.h
```diff
@@ -9,24 +9,34 @@ namespace ozo {
 struct connection_pool_config {
     std::size_t capacity = 10;
     std::size_t queue_capacity = 128;
+    time_traits::duration idle_timeout = std::chrono::seconds(60);
+};
+
+struct connection_pool_timeouts {
```
thed636: Could you please provide Doxygen compatible documentation about these fields.

thed636 (changes requested):

include/ozo/impl/async_request.h
elsid: Done

include/ozo/impl/async_request.h
elsid: Done

include/ozo/request.h
elsid: Done

include/ozo/impl/async_request.h
elsid: Done

include/ozo/impl/async_request.h
elsid: Done

include/ozo/impl/async_connect.h
elsid: `cancel` calls before timer setup, `done` after. Before we set timer there is no need to post client handler into strand.

include/ozo/impl/async_connect.h
elsid: Done

include/ozo/impl/async_connect.h
thed636: Is it bad to post handler into strand? I mean do we really need two different but very similar  functions here?

include/ozo/connection_pool.h
elsid: Done

include/ozo/impl/async_connect.h
elsid: Done

thed636 (approved):

PR #67, opened 2018-08-28
Asio executors

elsid:


include/ozo/impl/async_connect.h
```diff
@@ -102,10 +105,10 @@ struct async_connect_op {
         }
 
         get_timer(get_connection(context)).expires_after(timeout);
-        get_timer(get_connection(context)).async_wait(bind_executor(get_executor(context),
+        get_timer(get_connection(context)).async_wait(asio::bind_executor(get_executor(),
```
thed636, line 108: Is it possible to specify executor for the timeout handler via its constructor directly? IMO it would be better for the handler reusing.

thed636 (changes requested):

include/ozo/impl/async_connect.h
elsid: Yes, but what for? Better how?

thed636: Well, I think it will be useful e.g. for request time-out, and IMO it would be better to type something like this in both places:

```c++
get_timer(get_connection(context)).async_wait(
    make_timeout_handler(get_executor(), get_socket(get_connection(context)));
```

or even better may be you want to do something like this (may be now or may be in next PRs):
```c++
template <typename Conn, typename Executor> 
void set_cancel_timer(Conn& conn, Executor&& ex, const time_traits::duration& timeout) {
    decltype(auto) timer = get_timer(conn);
    decltype(auto) s = get_socket(conn);
    timer.expires_after(timeout);
    timer.async_wait(
        asio::bind_executor(std::forward<Executor>(ex), [s](error_code ec) {
            if (ec != asio::error::operation_aborted) {
                s.cancel(ec);
            }
        })
    );
}
```
and then it could be used like this:
```c++
set_cancel_timer(get_connection(context), get_executor(), timeout);
```

elsid: If we want refactor this place, we can do it later. This pr is about different thing.

thed636: Ok

thed636 (approved):

PR #68, opened 2018-08-30
Fix readme links

thed636:


README.md
```diff
@@ -12,8 +12,8 @@ The library leverages the power of template metaprogramming, providing convenien
 
 Since the project is on early state of development it lacks of documentation. We understand the importance of good docs and are working hard on this problem. Complete documentation is on the way, but now:
 
-* look at our brand new [HOW TO](docs/howto.md),
-* try our [generated from sources documentation](docs/html/index.html) made with [Doxypress](http://www.copperspice.com/documentation-doxypress.html) - it is under construction but readable,
+* look at our brand new [HOW TO](https://yandex.github.io/ozo/docs/howto.md),
```
elsid: This link will download file. Probably you wanted to put https://yandex.github.io/ozo/docs/html/md_docs_howto.html or just leave original, github can show md files.

elsid (changes requested):

README.md
thed636: The original link is not handled properly with Doxygen/Doxypress and disappears during processing.
I may change in this way - https://github.com/yandex/ozo/blob/master/docs/howto.md

elsid (approved):

PR #75, opened 2018-09-11
Add benchmarks

elsid:


include/ozo/type_traits.h
```diff
@@ -506,6 +510,7 @@ OZO_PG_DEFINE_TYPE_AND_ARRAY(float, "float4", FLOAT4OID, FLOAT4ARRAYOID, decltyp
 
 OZO_PG_DEFINE_TYPE_AND_ARRAY(std::string, "text", TEXTOID, TEXTARRAYOID, dynamic_size)
 
+OZO_PG_DEFINE_TYPE(ozo::name_oid, "nameoid", NAMEOID, dynamic_size)
```
thed636: It seems like this is a "name" type with an array OID = 1003. And it has `NAMEDATALEN` bytes.

    "63-byte type for storing system identifiers"

`NAMEDATALEN` is 64 bytes by default.

https://www.postgresql.org/docs/9.3/static/runtime-config-preset.html

But it seems like it is not good idea to define it as a separate type, since it can be changed while building PostgreSQL.

It is better to use macro for a full type definition:
   
    OZO_PG_DEFINE_TYPE_AND_ARRAY(ozo::pg::name, "name", NAMEOID, 1003, dynamic_size);

include/ozo/type_traits.h
```diff
@@ -403,6 +403,10 @@ constexpr auto size_of(const T& v) noexcept -> typename std::enable_if<
     return std::size(v) ? std::size(v) * size_of(*std::begin(v)) : 0;
 }
 
+struct name_oid {
+    std::string value;
+};
```
thed636: May be [BOOST_STRONG_TYPEDEF](https://www.boost.org/doc/libs/1_67_0/libs/serialization/doc/strong_typedef.html)? Or in case if the mechanism is not convenient for us make something similar?

E.g.

  In a new file with name `include/ozo/pg_types.h`

```cpp

#include <boost/serialization/strong_typedef.hpp>

namespace ozo::pg {
    BOOST_STRONG_TYPEDEF(std::string, name)
} // namespace ozo::pg

```

Or maybe make extend a macro like this:

```cpp

#define OZO_STRONG_TYPEDEF(BASE, NEW)
    BOOST_STRONG_TYPEDEF(BASE, NEW)
    BASE ozo_get_underlying_type(NEW&&);

```

So, in this case you can reuse send_impl and recv_impl for the type strong defined.

Or just copy and paste Boost solution and extend it with necessary methods and types.

thed636 (changes requested):

include/ozo/type_traits.h
elsid: Done

include/ozo/type_traits.h
elsid: Done

include/ozo/concept.h
```diff
@@ -253,7 +253,10 @@ template <typename T, typename = std::void_t<>>
 struct is_raw_data_writable : std::false_type {};
 
 template <typename T>
-struct is_raw_data_writable<T, std::void_t<decltype(std::declval<T&>().data())>> : std::true_type {};
+struct is_raw_data_writable<T, std::void_t<
+    decltype(std::declval<T&>().data()),
+    decltype(std::declval<T&>().size())
+>> : std::integral_constant<bool, std::is_same_v<decltype(std::declval<T&>().data()), char*>> {};
```
thed636, line 259: IMO instead of

```cpp
std::integral_constant<bool, std::is_same_v<decltype(std::declval<T&>().data()), char*>>
```

it is better to use 

```cpp
std::bool_constant<sizeof(decltype(*std::declval<T&>().data())) == 1>
```

Or even better - extend this concept like in this commit - https://github.com/yandex/ozo/pull/76/commits/8a3991863c62a08c6ee436c01512fe47c9b80829

thed636 (approved):

PR #83, opened 2018-10-19
Allow reference members in query parameters structure

IlyaSidorov:


thed636 (changes requested):
It would be nice to provide a test for a non default-constructible parameters for accompany this change.

thed636 (approved):

PR #85, opened 2018-10-22
Fix ussies with ozo::binder and ozo::connection_pool

thed636:
First - add allocator forwarding to ozo::binder
Second - add missing include of std::shared_ptr adaptation in the
connection pool implementation header.


thed636:
Duplicate

thed636:
Duplicate

PR #89, opened 2018-11-13
Async consume input after read rows

elsid:
When we have all data for rows, there is still could be some data in socket. To avoid synchronous read in `PQgetResult` we should poll socket until data is available and call `PQconsumeInput` to read from socket.

### DIFF
--- a/include/ozo/impl/async_request.h
+++ b/include/ozo/impl/async_request.h
@@ -22,6 +22,9 @@ struct request_operation_context {
     ozo::strand<decltype(get_io_context(conn))> strand {get_io_context(conn)};
     query_state state = query_state::send_in_progress;
 
+    using result_type = std::decay_t<decltype(get_result(conn))>;
+    result_type result;
+
     request_operation_context(Connection conn, Handler handler)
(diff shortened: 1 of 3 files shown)


include/ozo/impl/async_request.h
```diff
@@ -232,44 +246,52 @@ struct async_get_result_op : boost::asio::coroutine {
...
+            if (!get_request_result(ctx_)) {
+                return done();
             }
+
+            if (result_status(*get_request_result(ctx_)) != PGRES_SINGLE_TUPLE) {
+                do {
+                    while (is_busy(get_connection(ctx_))) {
+                        yield read_poll(ctx_, *this);
+                        if (auto err = consume_input(get_connection(ctx_))) {
+                            return done(err);
```
thed636: I'm not sure that we need to propagate error here since the result has been received successfully at the moment.

thed636 (changes requested):

include/ozo/impl/async_request.h
elsid: Fixed

thed636 (approved):

PR #93, opened 2018-11-19
Finite send query params

elsid:


include/ozo/impl/async_request.h
```diff
@@ -128,38 +128,63 @@ struct async_send_query_params_op {
...
         post(ctx_, *this);
     }
 
     void operator () (error_code ec = error_code{}, std::size_t = 0) {
-        // if data has been flushed or error has been set by
-        // read operation no write opertion handling is needed
-        // anymore.
-        if (get_query_state(ctx_) != query_state::send_in_progress) {
-            return;
+        switch (get_query_state(ctx_)) {
```
thed636, line 143: IMO it is better to place error check here instead of copy paste in two functions 

```c++
if (ec) {
    return done(ctx_, ec);
}
switch (get_query_state(ctx_)) {
```

include/ozo/impl/async_request.h
```diff
@@ -128,38 +128,63 @@ struct async_send_query_params_op {
...
         if (ec) {
             return done(ctx_, ec);
         }
 
         // Trying to flush output one more time according to the
         // documentation
         switch (flush_output(get_connection(ctx_))) {
-            case query_state::error:
+            case flush_result::error:
+                set_query_state(ctx_, query_state::error);
```
thed636, line 178: Setting the state into error is being made in `done()`. This call is not needed.

include/ozo/impl/async_request.h
```diff
@@ -128,38 +128,63 @@ struct async_send_query_params_op {
...
+        if (ec) {
+            return done(ctx_, ec);
+        }
+
+        if (send_query_params(get_connection(ctx_), query_)) {
+            set_query_state(ctx_, query_state::flushing);
+            (*this)();
+        } else {
+            post(ctx_, *this);
+        }
```
thed636, line 166: This function can be reused in `perform()` if it always use `post()` and do not call `operator()` directly. Or you can try to leave only `post()` call inside `perform()` and call `send_query_params()` only here.

thed636 (changes requested):

thed636:
It seems like the retries aren't needed

https://github.com/yandex/ozo/commit/12441cd3c0188944a65340a25a5ccaccdc5d1117

PR #94, opened 2018-11-20
Support std::string_view for params type

elsid:


tests/binary_serialization.cpp
```diff
@@ -64,6 +66,12 @@ TEST_F(send, with_std_string_should_store_it_as_is) {
     EXPECT_THAT(buffer, ElementsAre('t', 'e', 'x', 't'));
 }
 
+TEST_F(send, with_std_string_view_should_store_it_as_is) {
```
thed636, line 69: We can not receive into `std::string_view`, so it would be nice to have the good readable error message about it if somebody tries to receive into `std::string_view`.  It seems like the error message can be triggered from code around `recv_impl` or through the template specialisation.

thed636 (changes requested):

tests/binary_serialization.cpp
elsid: I've tried to generalize check for type to be sendable or receivable. So a client should have an error by static_assert. There are a lot of cases what can go wrong with user type. I've added to an error message only for most probable.

thed636 (approved):

PR #99, opened 2018-12-11
Support conversion to bool for transaction

elsid:


include/ozo/impl/transaction.h
```diff
@@ -27,6 +27,14 @@ class transaction {
...
 
+    bool has_connection() const {
+        return impl != nullptr && impl->has_connection();
+    }
+
+    operator bool() const {
+        return has_connection();
+    }
+
     friend auto& unwrap_connection(transaction& self) noexcept {
```
thed636: If it possible, could you please to change unwrap_connection customisations to the structure-based  in this pull request?

```cpp
namespace ozo {

template <typename T>
struct unwrap_connection_impl<impl::transaction<T>> {
    template <typename Transaction>
    static constexpr decltype(auto) apply(Transaction&& v) noexcept {
        return unwrap_connection(*v.impl->connection);
    }
};

} // namespace zoo
```

elsid: This wouldn't work as is, because `impl` is a private field. I still need to use a friend function or class.

thed636: Well, in this case friendship of `unwrap_connection_impl` is preferable.

elsid: Done

thed636 (approved):

PR #104, opened 2018-12-27
Fix size_of() for strong_typedef_wrapper wrapped types

thed636:
The problem is what the previous fix is not correct. It leads to application
of size calculation algorithm for arrays instead of default algorithm for
dynamic size data.

New test are added to address this problem.

### DIFF
--- a/tests/binary_serialization.cpp
+++ b/tests/binary_serialization.cpp
@@ -85,11 +85,6 @@ TEST_F(send, with_std_vector_of_float_should_store_with_one_dimension_array_head
     }));
 }
 
-TEST_F(send, should_convert_pg_name_as_string) {
-    ozo::send(os, oid_map, ozo::pg::name {"name"});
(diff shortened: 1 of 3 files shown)


tests/binary_serialization.cpp
```diff
@@ -120,4 +115,44 @@ TEST(send_impl, should_send_nothing_for_std_nullopt_t) {
...
+    auto size_buffer() {
+        return buffer_range(4, 4);
+    }
+
+    auto data_buffer() {
+        return boost::make_iterator_range(buffer.begin() + 8, buffer.end());
+    }
+};
+
+TEST_F(send_frame, should_write_pg_bytea_as_bunary_byte_buffer) {
```
elsid: binary?

elsid (changes requested):

tests/binary_serialization.cpp
thed636: **Done**

elsid (approved):

PR #106, opened 2019-01-09
Don't apply -1 length to offset

elsid:
This length is reserved for `NULL` values.

### DIFF
--- a/include/ozo/io/binary_query.h
+++ b/include/ozo/io/binary_query.h
@@ -134,8 +134,12 @@ private:
         std::size_t offset = 0;
         hana::for_each(range, [&] (auto field) {
                 const auto size = result->lengths[field];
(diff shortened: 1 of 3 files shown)


include/ozo/io/binary_query.h
```diff
@@ -134,8 +134,12 @@ class binary_query {
         std::size_t offset = 0;
         hana::for_each(range, [&] (auto field) {
                 const auto size = result->lengths[field];
-                result->values[field] = size ? std::data(result->buffer) + offset : nullptr;
-                offset += size;
+                if (size && size != -1) {
```
thed636: There is special constant type for it
```suggestion
                if (size && size != null_state_size) {
```

thed636 (changes requested):

thed636 (dismissed):

thed636 (approved):

PR #114, opened 2019-02-14
Add default constructor to query_repository

elsid:


include/ozo/query_conf.h
```diff
@@ -418,9 +418,15 @@ inline std::shared_ptr<query_conf> make_query_conf(std::vector<ozo::detail::quer
 template <class ... QueriesT>
 class query_repository {
 public:
+    query_repository() = default;
+
     query_repository(std::shared_ptr<detail::query_conf> query_conf)
         : query_conf(std::move(query_conf)) {}
 
+    bool has_query_conf() const {
```
thed636: 1. IMO it is better to give more common name like `is_initialized()` or something like this. Analog to std::optional's `has_value()` in this case is not suitable due to this class does not provide any query_conf object - uses it but does not provide.
2. It would be nice to provide an overloaded `operator bool()` for convenience as well.

thed636 (changes requested):

include/ozo/query_conf.h
elsid: Done

thed636 (approved):

PR #125, opened 2019-04-01
[INCOMPATIBLE] unwrap() and Nullable  refactoring

thed636:
Refactoring of the unwrap() function and Nullable concept usage. This refactoring is dedicated to the decoupling of these two. This PR may break user code in rare cases - the code won't compile if a user expects the dereferencing on calling unwrap() with custom Nullable type argument with no unwrap_impl specialization.


massaraksh111:
As for me it will be better to unit-test new unwrap logic

thed636:
> As for me it will be better to unit-test new unwrap logic

Yeah, thanks, sure.

massaraksh111 (approved):

PR #129, opened 2019-05-16
Failover framework

thed636:
Failover micro framework
===========

Introduce failover mechanism with retry strategy as a first implemented strategy.

RATIONALE
----------

Distributed environment is unstable, so we want to have recovery mechanisms. Failover micro-framework is created to provide such customisable and easy to use solution. Here we have two new concepts: *Strategy* and *Try*.

* **Strategy** is entity which determines how to try to recover an operation which failed. 
* **Try** is an entity which holds a context of an operation and decides should we try to recover a failed operation or give an error for a user.

EXTENSION
----------

The `ozo::fallback::retry` strategy is an example of the framework extension.

OPERATION SUPPORT FAILOVER
-----------------

* request
* execute
* get_connection
* begin

BASIC USAGE EXAMPLE
----------------

Operation with failover looks like this:

```c++
execute[fallback::retry(connection_error)*3](conn_info[io], query, 3s, yield);
```

see [retry_request.cpp](https://github.com/yandex/ozo/pull/129/files#diff-0faf5b07ab5932bd18487a1b85009ef6) for more details.

### DIFF
--- a/include/ozo/failover/retry.h
+++ b/include/ozo/failover/retry.h
@@ -1,0 +1,286 @@
+#pragma once
+
+#include <ozo/failover/strategy.h>
+#include <boost/hana/is_empty.hpp>
+#include <boost/hana/tuple.hpp>
+#include <boost/hana/fold.hpp>
+#include <boost/hana/concat.hpp>
+
+/**
+ * @defgroup group-failover-retry Retry
+ * @ingroup group-failover
+ * @brief Failover by simple retry operation
+ */
+
+namespace ozo::failover {
+
+namespace detail {
+
+template <typename TimeConstraint, typename Now = decltype(time_traits::now)>
+inline auto get_try_time_constraint(TimeConstraint t, int n_tries, [[maybe_unused]] Now now = time_traits::now) {
+    if constexpr (t == none) {
+        return none;
+    } else if constexpr (std::is_same_v<TimeConstraint, time_traits::time_point>) {
+        return n_tries > 0 ? time_left(t, now()) / n_tries : time_traits::duration{0};
+    } else {
+        return n_tries > 0 ? t / n_tries : time_traits::duration{0};
+    }
+}
+
+template <typename ...ErrorConditions>
+constexpr auto make_errcs_tuple(ErrorConditions ...errcs) {
+    return hana::make_tuple(errcs...);
+}
+
+} // namespace detail
+
+/**
+ * @brief Operation try representation
+ *
+ * Controls current try state incluiding number of tries remain, context of operation call for
+ * current try.
+ *
+ * @tparam ErrorConditions --- error conditions to match on for retry, should model #HanaSequence
+ * @tparam Context --- operation call context, literally operation arguments except #CompletionToken
+ * @ingroup group-failover-retry
+ */
+template <typename ErrorConditions, typename Context>
+class basic_try {
+public:
+    static_assert(HanaSequence<ErrorConditions>, "ErrorConditions should models HanaSequence");
+
+    /**
+     * @brief Construct a new basic try object.
+     *
+     * @param n_tries --- number of tries allowed, should not be less than 0.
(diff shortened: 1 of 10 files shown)


include/ozo/detail/cancel_timer_handler.h
```diff
@@ -30,4 +31,25 @@ struct cancel_timer_handler {
...
+
+template <>
+struct bind_cancel_timer_impl<ozo::none_t> {
+    template <typename Handler>
+    constexpr static decltype(auto) apply(Handler&& h) {
+        return std::forward<Handler>(h);
+    }
+};
+
+template <typename TimeConstrain, typename Handler>
```
yagzalex: Did you mean `TimeConstraint`?

thed636: Yep, thanks.

include/ozo/failover/retry.h
```diff
@@ -0,0 +1,281 @@
...
+     * - for the 1st try: 0,5/3 sec.
+     * - for the 2nd try: (0,5 - t<small>1</small>) / 2 >= 0,5/3 sec.
+     * - for the 3rd try: 0,5 - (t<small>1</small> + t<small>2</small>) >= 0,5/3 sec.
+     *
+     * @code
+    auto retry = failover::retry(errc::network, errc::timeout);
+    ozo::request[retry*3](pool, query, .5s, out, yield);
+     * @endcode
+     */
+    auto times(int n) const {
```
yagzalex: Maybe you should named that as 'tries'. The name 'times' is really confusing, cause it can be misunderstood as time constraints or something.

tests/failover/retry.cpp
```diff
@@ -0,0 +1,191 @@
...
+TEST_F(get_try_time_constraint, should_return_zero_duration_for_try_count_less_than_zero) {
+    EXPECT_EQ(duration{0}, ozo::failover::detail::get_try_time_constraint(3s, -1));
+}
+
+TEST_F(get_try_time_constraint, should_return_time_left_divided_on_try_count_for_try_count_greter_than_zero) {
+    EXPECT_EQ(1s, ozo::failover::detail::get_try_time_constraint(deadline, 3, now));
+}
+
+TEST_F(get_try_time_constraint, should_return_zero_time_left_for_try_count_zero) {
+    EXPECT_EQ(duration{0}, ozo::failover::detail::get_try_time_constraint(3s, 0, now));
```
yagzalex: Don't you mean to use 'deadline' from here? Otherwise, I don't see a big difference with 'should_return_zero_duration_for_try_count_zero' test.

include/ozo/failover/retry.h
```diff
@@ -0,0 +1,281 @@
...
+        return retval;
+    }
+
+    /**
+     * @brief Sintactic sugar for `ozo::retry_strategy::times()`
+     *
+     * @param n --- number of tries
+     * @return `retry_strategy` specialization object
+     */
+    auto operator * (int n) const { return times(n); }
```
yagzalex: Crazy thought. I think it would be handy to also have
```
template <typename ErrorConditions>
auto operator * (int n, retry_strategy<ErrorConditions> rs);
```
Because when you write `ozo::request[ozo::failover::retry(ozo::errc::network) * 3](...)` it is a little bit hard to find the exact counts of tries.
And this will make it easier `ozo::request[3 * ozo::failover::retry(ozo::errc::network)](...)`

yagzalex (changes requested):

include/ozo/failover/retry.h
thed636: **Done**

tests/failover/retry.cpp
thed636: **Done**
Yeah, thanks!

include/ozo/failover/retry.h
thed636: **Done**

yagzalex (approved):

Issue #132, opened 2019-05-18
How to retrieve RAISEd error messages


JonasProgrammer:
First of all, thank you for this awesome library!
I currently have an issue with getting the cause of a transaction abort within an `ozo` handler. Neither `ec.message()` nor `ozo::get_error_context` seem to do the job, however, as I just get a generic 'transaction aborted' message in both.

Basically I have something like this:
```sql
CREATE FUNCTION foo(bar int) [...]
IF bar = 42 THEN RAISE '$err1$ some message' END IF;
IF bar < 0 THEN RAISE '$err2$ other message' END IF;
[...]
```

And I'm trying to achieve this:
```cpp
ozo::begin(/*...*/, [](ozo::error_code ec, auto conn) {
  if (ec) return;
  ozo::execute(std::move(conn), "SELECT foo("_SQL + 42 + ")"_SQL, [](ozo::error_code ec, auto conn) {
    if (ec)
    {
        auto err = ec.message(); // ozo::get_error_context(conn) doesn't work either
        // if err contains $err1$ do one thing, do something else otherwise
// ...
```

Please beware that I'm using a forked version of ozo that works with boost 1.70 (and uses targets in CMake, but that shouldn't change the behaviour of a header only lib), so if this approach should work with vanilla ozo, please tell me, so I can look where I screwed up.

thed636:
Hi Jonas!

Thanks for the issue.

Try to get native error message via [ozo::error_message()](https://yandex.github.io/ozo/group__group-connection-functions.html#gae1d570854c32a38adcad99baf73e0baa) function. Unfortunately it is not very clear since the function is not used in any example - let's do not close this issue until I've change the examples.

BTW is it possible to see the fork of the project? I'm interesting in both changes, so if this fork is yours then it would be nice to have a PR in the project.

JonasProgrammer:
@thed636 Thanks alot, you're saving my weekend. `ozo::error_message()` works like a charm.

[Here's the fork](https://github.com/JonasProgrammer/ozo/tree/boost-170-ozo) (the branch name should have been `boost-170-conan`).

I thought about PRing, but I didn't, because this is basically what should have been 3 feature branches. At the same time this:
1. changes basically all CMake files to use targets (and breaks Coverage in doing so, as I didn't need it and couldn't be bothered to port it yet)
2. adds a conan recipe packing the 'targetized' version for CMake
[...]

PR #133, opened 2019-05-19
Add more documentation about error handling

thed636:
Add ozo::error_message() into example code.

Addresses issue #132

### DIFF
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -2,6 +2,14 @@
 
 Here are some examples of how to use OZO API.
 
+<!-- TOC -->
+- [How to](#how-to)
+  - [How To Make A Very Simple request](#how-to-make-a-very-simple-request)
+  - [How To Handle Error Properly](#how-to-handle-error-properly)
+  - [How Start To Do Not Bother With Types Sequence In Row But Begin To Bother About Column Names](#how-start-to-do-not-bother-with-types-sequence-in-row-but-begin-to-bother-about-column-names)
+  - [How To Determine Which Type Do I Need To Use For The PostgreSQL Type](#how-to-determine-which-type-do-i-need-to-use-for-the-postgresql-type)
+  - [How To Bind One More PostgreSQL Type For C++ Type With Existing Binding](#how-to-bind-one-more-postgresql-type-for-c-type-with-existing-binding)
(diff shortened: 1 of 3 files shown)


docs/howto.md
```diff
@@ -39,63 +47,63 @@ int main() {
...
 
-Now we need to create a connection information for database to connect to. This is our connection provider which can create a connection for us as it will be needed (see more information about connection provider and how to get connection).
+Now we need to create a connection information for database to connect to. This is our connection source which can create a connection for us as it will be needed (you can learn more about ConnectionSource and ConnectionProvider concepts from the documentation).
 
 ```cpp
 const auto query = "SELECT id, name FROM users_info WHERE amount>="_SQL + std::int64_t(25);
 ```
 
-Here our query for database. There is a text with `_SQL` literal and single parameter `std::int64_t(25)`. Note, what the parameter will be passed as a separate binary parameter but not as a part of query text.
+Here our query for database. There is a text with `_SQL` literal and a parameter `std::int64_t(25)`. Note, what the parameter will be passed as a separate binary parameter but not as a part of query text.
```
yagzalex: Note that the parameter will be passed as a separate binary parameter, but not as part of the query text.

docs/howto.md
```diff
@@ -106,16 +114,49 @@ ozo::request(ozo::make_connector(conn_info, io), query, ozo::into(res),
...
 
 `for(auto& row: res)` - so if there is no error we can handle result from the output container.
 
+---
+
+## How To Handle Error Properly
+
+OZO uses `boost::system::error_code` to indicate error and so as `std::error_code` too it can not provide any context depended information at all. So unfortunately `error_code::message()` returns the only a static textual description of the code. But this is not enough, especially for sql errors. That's why the additional error information is needed and can be obtained via these two functions:
```
yagzalex: OZO uses `boost::system::error_code` to indicate an error and therefore, like` std::error_code`, it cannot provide any context-dependent information at all. So unfortunately `error_code::message()` returns only a static textual description of the code.

yagzalex (commented):
Maybe these wordings will work better.

yagzalex (approved):

PR #136, opened 2019-05-20
Boost 1.70 support

### LINKED ISSUE #134: Boos 1.70 compatibility
Currently OZO does not support Boost 1.70 and this is no good.

JonasProgrammer:
Fixups for Boost 1.70 (which are downwards compatible to boost 1.67):
* updated `resourece_pool` (likely has to be re-based back to the ancestor repository)
* use `get_executor().context()` instead of deprecated/removed `get_io_context()`
* fix `test_asio.h` to always have an executor to make io context retrieval work

Fixes #134

### DIFF
--- a/include/ozo/connection.h
+++ b/include/ozo/connection.h
@@ -434,7 +434,7 @@ inline decltype(auto) get_socket(T&& conn) noexcept {
 template <typename T>
 inline decltype(auto) get_io_context(T& conn) noexcept {
     static_assert(Connection<T>, "T must be a Connection");
-    return get_socket(conn).get_io_context();
+    return get_socket(conn).get_executor().context();
 }
 
 /**
--- a/tests/test_asio.h
+++ b/tests/test_asio.h
@@ -59,6 +59,9 @@ struct execution_context : asio::execution_context {
         executor_type(executor_mock& impl, execution_context& context)
         : impl_(&impl), context_(&context){}
 
+        explicit executor_type(execution_context& context)
+        : context_(&context){}
+
         execution_context& context() const noexcept {
             return *context_;
         }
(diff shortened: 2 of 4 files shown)


include/ozo/connection.h
```diff
@@ -433,7 +433,11 @@ inline decltype(auto) get_socket(T&& conn) noexcept {
 template <typename T>
 inline decltype(auto) get_io_context(T& conn) noexcept {
     static_assert(Connection<T>, "T must be a Connection");
+#if BOOST_VERSION < 106700
     return get_socket(conn).get_io_context();
+#else
+    return get_socket(conn).get_executor().context();
+#endif
```
thed636: Looks like this condition is not needed, because in Boost 1.66  `boost::asioposix::stream_descriptor::get_executor()` [already exists](https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/posix__basic_stream_descriptor/get_executor.html), so it can be used safely.

tests/test_asio.h
```diff
@@ -69,19 +72,27 @@ struct execution_context : asio::execution_context {
...
 
         template <typename Function>
         void defer(Function&& f, std::allocator<void>) const {
+            assert_has_impl();
             return impl_->defer(wrap_shared(std::forward<Function>(f)));
         }
 
+        void assert_has_impl() const {
+            if (!impl_)
+                throw std::logic_error("ozo::testing::execution_context::executor_type::assert_impl() no executor mock");
```
thed636, line 93: Could you please place curly braces for if statement, like in project. The rationale for this are:
- changes inside `if ()` statement scope will not affect statement itself, so the history of the project will be more clear,
- sometimes weird  things like `if();` and similar are happened, so the braces help to avoid such thing.

thed636 (changes requested):
Thanks for the PR!

To make your first PR legally acceptable it is needed to add you into the [AUTHORS](https://github.com/yandex/ozo/blob/master/AUTHORS). The string with your name and e-mail would mean that you're agree with text on first two strings. It is necessary since the project is in the Yandex's  organisation.

thed636 (dismissed):

thed636 (dismissed):

thed636 (approved):

PR #139, opened 2019-06-01
Add cancel() operation for database request

thed636:
CANCEL OPERATION
===============

Sometimes you need to cancel request to a database due to operation time constraint. In case of closing connection, PostgreSQL backend will continue to execute request anyway. So to prevent such waste of database resources it is best practice to cancel the execution of the request by sending special command.

This PR enables such a feature via a standard underlying libpq function.

ASYNCHRONOUS CALL PROBLEM
========================

Since libpq's cancel operation implementation is synchronous only, the default behaviour is to execute the operation via `boost::asio::system_executor` in a separate thread for each call. User may specify own executor, to implement different strategy, e.g. cancel operation queue which would be handled in dedicated thread and so on.

The other solution is own asynchronous implementation of cancel protocol. For now this solution is less attractive because in long term it leads to tracking of protocol changes. I'd like to avoid such things.


thed636:
Updated for operation functors, but still needs integration tests. The only solution I see try to implement single row result mode and cancel the request between getting rows or something like this.

yagzalex (approved):

PR #141, opened 2019-06-07
Async initiate & error conditions

thed636:
Support async_initiate
=========

In Boost 1.70 Asio introduces a new version of the universal asynchronous interface which supports Coroutine TS. Fortunately this mechanism become a good basis for the failover framework via initiators. Initiator is a customisation point which allows to change execution strategy from straight forward to any of failover strategies.

Error Conditions
============

Set-up error conditions of the `sqlstate` error classes - now classes error conditions match to all of their errors. Introduce library error conditions `ozo::errc::code` to be useful in matching to failover conditions.


include/ozo/error.h
```diff
@@ -349,30 +354,59 @@ enum code {
...
 enum code {
-    ok,
-    sql_error,
-    connect_error,
-    transport_error,
-    database_readonly,
+    ok, //!< no error placeholder
+    connection_error, //!< connection-related error condition, incorporates ozo, libpq and Boost.Asio connection errors
+    database_readonly, //!< database in read-only state - useful to detect attempt of modify data on replica host
+    introspection_error, //<! errors related to objects serealisation/deserialization
```
dpotepalov: serealization -> serialization

include/ozo/error.h
```diff
@@ -731,4 +822,141 @@ inline const error_category& category() noexcept {
...
+        ozo::error::bad_object_size,
+        ozo::error::bad_array_size,
+        ozo::error::bad_array_dimension,
+        ozo::error::bad_composite_size,
+        ozo::error::unexpected_eof
+    );
+};
+
+template <>
+struct codes_for_condition<type_mismatch> {
```
dpotepalov, line 895: Why `type_mismatch` is not an introspection error? It fits into the category "could not translate the bytes DB sent us into an instance of user-supplied type", does it not?

thed636: Type mismatch could occur in case of database schema changed and the types get new OIDs, so in this case, simple reconnect should help and this error condition can be used to detect such a situation.

include/ozo/error.h
thed636: **Done**

yagzalex (approved):

PR #143, opened 2019-06-10
Improve CMake integration

omartijn:
* Now provides target 'ozo', which automatically pulls in dependencies
* Package can now be found using add_subdirectory or find_package
* Now also adds an install option, which installs headers and CMake files

It's not perfect yet, especially since the resource_pool package also doesn't provide a proper target, but IMHO it's a step in the right direction.


thed636:
@omartijn thanks for the PR, well it seems like it needs a little bit more tuned because it passes no CI tests.

@JonasProgrammer , could you please to look at this PR? It seems like it very close to your target-based cmake.

omartijn:
Ah, I see it now. It seems this is due to a some changes made in boost, which break dependency resolution for components in older versions of CMake. Basically, if you do find_package() to find a boost component, like boost::coroutine, all the necessary dependencies for it (like boost::system and boost::thread) should also be found automatically.

This is probably why these components were specifically mentioned in the CMakeLists.txt, even though they shouldn't be necessary. I'll put them back with a comment explaining why they're there (so nobody else will try removing them).

omartijn:
I can't get the tests to work, because for some reason it wants to have boost::atomic, which is not built because the `--with-libraries` option doesn't contain atomic.

omartijn:
I updated the Dockerfile as well to also build boost::atomic, and now everything builds, except osx, but that seems to be more of an issue with using a gnu-specific feature in the code.

JonasProgrammer:
@thed636 Yep, I did something similar, the exception being I kept it compatible with the current build system, unless `OZO_MODERN_CMAKE` was defined.

Depending on how high you can go in regards to the cmake version in Yandex's ecosystem, I'd like to suggest the following improvements:
*  Use a more recent CMake version, that supports `target_compile_features` etc. -- the library can state it wants at least `cxx_std_17`
* Use `target_link_libraries` with targets like `Boost::context` instead of `${Boost_LIBRARIES}` and `${PostgreSQL_LIBRARIES}`
* At least the tests still use `CMAKE_CXX_FLAGS`. While they're considered 'private', I still think these could be updated to use `target_compile_options`, when we're already at it.
* Perhaps GMock and GTest could be imported via targets as well (with hints to the vendored directories), rather than hard-including them.
* `resource_pool` should definitely benefit from this effort. I can try and polish up my CMake files to make it PR-worthy, but it won't be before the weekend. Regarding this: @elsid Do you have any dependent projects relying on a cmake 2.x compatible library being present? If not, I could skip the compatibility part.

@omartijn  These are suggestions only and up for discussion. Thanks for putting so much effort into it.

EDIT: While we're at it, what's Yandex's general attitude towards package managers like `conan`? If we get the target-based libraries with xConfig.cmake files in, as suggested by this PR, it's merely adding one file to the repo and it's use is purely optional, but makes it much easier to get `ozo` up and running for a project without going through the hassle of setting up Boost and libpq manually.

omartijn:
> @thed636 Yep, I did something similar, the exception being I kept it compatible with the current build system, unless `OZO_MODERN_CMAKE` was defined.
> 
> Depending on how high you can go in regards to the cmake version in Yandex's ecosystem, I'd like to suggest the following improvements:
> 
>     * Use a more recent CMake version, that supports `target_compile_features` etc. -- the library can state it wants at least `cxx_std_17`

That's introduced in 3.1.3. That'd be fine for me, but I have no idea if that works for everybody. Perhaps somebody can tell me what the minimum required version has to be.

> 
>     * Use `target_link_libraries` with targets like `Boost::context` instead of `${Boost_LIBRARIES}` and `${PostgreSQL_LIBRARIES}`

Yup. I did this for boost already. The `FindPostgreSQL.cmake` included with my version of cmake (3.13.4) doesn't support imported targets for PostgreSQL, however, so this doesn't work for me. I could of course include this file in the repo as well (it's right there in kitware's repo). Then we'd be able to support this for everyone.

> 
>     * At least the tests still use `CMAKE_CXX_FLAGS`. While they're considered 'private', I still think these could be updated to use `target_compile_options`, when we're already at it.

I didn't do this, for the reasons you mentioned. It doesn't matter in regards to how you use the `ozo` target. I also didn't want to put too much time in it before I knew if there was any interest at all. 

> 
>     * Perhaps GMock and GTest could be imported via targets as well (with hints to the vendored directories), rather than hard-including them.

I'm not so sure whether that's a good idea. I know in the past gtest was included in debian, but Google actually recommends against this now, see https://github.com/doo/googletest/blob/master/wiki/FAQ.wiki#Why_is_it_not_recommended_to_install_a_precompiled_copy_of_Google_Test_for_example_into_usrlocal

> 
>     * `resource_pool` should definitely benefit from this effort. I can try and polish up my CMake files to make it PR-worthy, but it won't be before the weekend. Regarding this: @elsid Do you have any dependent projects relying on a cmake 2.x compatible library being present? If not, I could skip the compatibility part.

That'd be great. Do you have a link to the existing work you did?

> 
> @omartijn These are suggestions only and up for discussion. Thanks for putting so much effort into it.
> 
> EDIT: While we're at it, what's Yandex's general attitude towards package managers like `conan`? If we get the target-based libraries with xConfig.cmake files in, as suggested by this PR, it's merely adding one file to the repo and it's use is purely optional, but makes it much easier to get `ozo` up and running for a project without going through the hassle of setting up Boost and libpq manually.

JonasProgrammer:
> > @thed636 Yep, I did something similar, the exception being I kept it compatible with the current build system, unless `OZO_MODERN_CMAKE` was defined.
> > Depending on how high you can go in regards to the cmake version in Yandex's ecosystem, I'd like to suggest the following improvements:
> > ```
> > * Use a more recent CMake version, that supports `target_compile_features` etc. -- the library can state it wants at least `cxx_std_17`
> > ```
> 
> That's introduced in 3.1.3. That'd be fine for me, but I have no idea if that works for everybody. Perhaps somebody can tell me what the minimum required version has to be.
> 
Yep, finding a common denominator would be the best thing to know what we can and cannot use.
> > ```
> > * Use `target_link_libraries` with targets like `Boost::context` instead of `${Boost_LIBRARIES}` and `${PostgreSQL_LIBRARIES}`
> > ```
> 
> Yup. I did this for boost already. The `FindPostgreSQL.cmake` included with my version of cmake (3.13.4) doesn't support imported targets for PostgreSQL, however, so this doesn't work for me. I could of course include this file in the repo as well (it's right there in kitware's repo). Then we'd be able to support this for everyone.
>
I tested it with 3.14 and it worked, but that might have been Conan magic. I like your approach of including it, that should do, till the minimum required CMake version is raised some day.
> > ```
> > * At least the tests still use `CMAKE_CXX_FLAGS`. While they're considered 'private', I still think these could be updated to use `target_compile_options`, when we're already at it.
> > ```
> 
> I didn't do this, for the reasons you mentioned. It doesn't matter in regards to how you use the `ozo` target. I also didn't want to put too much time in it before I knew if there was any interest at all.
> 
Understandable, I just saw it and remembered changing that. Still, as there seems to be interest in the PR, this shouldn't be too hard to change.
> > ```
> > * Perhaps GMock and GTest could be imported via targets as well (with hints to the vendored directories), rather than hard-including them.
> > ```
> 
> I'm not so sure whether that's a good idea. I know in the past gtest was included in debian, but Google actually recommends against this now, see https://github.com/doo/googletest/blob/master/wiki/FAQ.wiki#Why_is_it_not_recommended_to_install_a_precompiled_copy_of_Google_Test_for_example_into_usrlocal
> 
Reading that, I fully agree with you. For me it just instinctively looked clearer this way; let's just take that, as it is.
> > ```
> > * `resource_pool` should definitely benefit from this effort. I can try and polish up my CMake files to make it PR-worthy, but it won't be before the weekend. Regarding this: @elsid Do you have any dependent projects relying on a cmake 2.x compatible library being present? If not, I could skip the compatibility part.
> > ```
> 
> That'd be great. Do you have a link to the existing work you did?
> 

[here](https://github.com/JonasProgrammer/ozo/tree/boost-170-ozo)
Beware it's a little behind master and also it's weird in parts, as it tries to support the old build system as-is. Maybe you can still use it for something, such as resource_pool (though upstream PR would likely be the better option here).

omartijn:
@thed636 Any updates on this? Me and @JonasProgrammer would still like to know what the minimum required CMake version should be. If the one I selected is acceptable, I think this PR could be merged.

If we do the same target-based strategy for resource_pool as well, we'd need a few more changes to the CMakeLists.txt as well, but nothing major. Is there anything holding a possible review back right now?

thed636:
@omartijn @JonasProgrammer  
Hi guys! 
We are using own build toolchain inside the company instead of the CMake. So from the company, we have no restriction on the CMake version. In terms of Open Source users - I didn't think about the proper minimum version the project should require. So I need a couple of days to review this PR (busy week) and just form my opinion about the minimum version.

Thank you!

**BTW**, the MacOS build is still broken and it needs to be fixed anyway.

omartijn:
@thed636 Looking at CMake releases, e.g. at https://cmake.org/files/v3.1/ we can see that 3.1.3 was released 2015-02-12 17:39, which is well over 4 years ago. I personally think it's not unreasonable to expect people to have that version at the very least.

JonasProgrammer:
While thed636 obviously has the final say in this, IMHO even demanding the latest CMake version should not be too much of an issue.
First of all, the C++ ecosystem has become ever faster in the recent years. I therefore expect someone, who has access to boost 1.67+ and a recent compiler to also have access to a recent CMake version.
Ok, one could argue that Debian stable etc. don't provide a recent version, but neither do they provide a boost version and compilers (?) capable of dealing with ozo.

Another fact to look at, that all we're talking about here is the CMake version required to build this project. When it comes to the files being generated for `find_package`, they should be compatible with earlier CMake versions as well. Checking the `ozoTargets.cmake` generated by my changes using CMake 3.14, they even do some checks regarding compatibility with CMake >= 2.6.

Therefore IMHO we should just use a recent version. If somebody wants to build it from scratch, they can just use a recent CMake version and for the rest, adding some releases with the generated files used by `find_package` should be sufficient; this process can easily be automated with travis.

thed636:
@omartijn @JonasProgrammer 
Hi, guys! 

Thanks a lot for the patience. I'm afraid we are limited with [3.12 version](https://launchpad.net/ubuntu/cosmic/+source/cmake) right now due to our build image. Changing image is out of this PR. So, my proposal is to implement the best what the 3.12 version allows for now. Then create two issues. First for upgrade image to support CMake 3.14, second - upgrade CMake to 3.14 and update the cmake files adding the functionality is not supported by the lower version. Sounds good?

> First of all, the C++ ecosystem has become ever faster in the recent years. I therefore expect someone, who has access to boost 1.67+ and a recent compiler to also have access to a recent CMake version.

I'd say, we should support Boost 1.66+ as declared.

omartijn:
@thed636 I'll see what things I can make nicer using CMake 3.12. About changing to 3.14: I think the only improvement there is that the FindPostgreSQL.cmake file comes with imported targets, but I already work around it by including the file in the repository. If it doesn't have any other features besides that, I see no reason to require 3.14, it'd only make it more difficult for users.

There was also a question about supporting conan. @JonasProgrammer has some ideas about that. I have never worked with it, but if it's a simple line added to the CMakeLists.txt - and it doesn't require somebody to actually have conan installed for building, I'd think it should be no problem to add it as well, right? If we are doing this, we might as well add support for building debian and rpm packages. CMake has built-in support for this and we'd again make it easier for people to install and use the library.

Any ideas about this?

JonasProgrammer:
Hi everyone.

> > First of all, the C++ ecosystem has become ever faster in the recent years. I therefore expect someone, who has access to boost 1.67+ and a recent compiler to also have access to a recent CMake version.
> 
> I'd say, we should support Boost 1.66+ as declared.

My bad, I actually meant 1.66. Still, the point holds: Debian stable is at 1.62, Ubuntu LTS (Bionic) at 1.65. No idea about CentOS & co., but I doubt, it's much better there. In the end, to use this library, you cannot rely on Boost being provided by your system package manager. Downloading a more recent CMake version doesn't make matters much worse.

> Thanks a lot for the patience. I'm afraid we are limited with [3.12 version](https://launchpad.net/ubuntu/cosmic/+source/cmake) right now due to our build image. Changing image is out of this PR. So, my proposal is to implement the best what the 3.12 version allows for now.

I don't see an issue with that. CMake 3.12 is decent enough, the reason I was using 3.14 in my fork was simply that it was my build image's version.
Judging from the Changelog, there was not much relevant going on from 3.12 > 3.13 > 3.14 and all the good stuff for the target based approach is present in 3.12. So if that's yandex's baseline, let's just settle on this.

Also I'm with @omartijn regarding an update to 3.14 in the near future: The benefits don't outweigh the costs. As long as there's no really beneficial changes (like targets were), upgrading is not necessary. I really do appreciate your willingness to upgrade though; the 'we've always done it like this'-disease is way too common, when it comes to libraries' build systems.


> There was also a question about supporting conan. @JonasProgrammer has some ideas about that. I have never worked with it, but if it's a simple line added to the CMakeLists.txt - and it doesn't require somebody to actually have conan installed for building, I'd think it should be no problem to add it as well, right? If we are doing this, we might as well add support for building debian and rpm packages. CMake has built-in support for this and we'd again make it easier for people to install and use the library.
> 
> Any ideas about this?

Actually, the 3 lines in CMakeLists.txt [are necessary not even necessary](https://docs.conan.io/en/latest/integrations/build_system/cmake/cmake_paths_generator.html) -- they're just a convenience, because I'm way too lazy. You can easily specify the file in the CMake command line, if you generated it.
On the other hand, having them doesn't really hurt, as the build process just behaves as if they weren't there, given the file does not exist. In general, including them or not is up to you.

The real deal is providing a `conanfile.py` and perhaps a test package -- but given a proper `CMakeLists.txt`, as is provided by this PR, this is not a big deal either. I can try and incorporate them here, or just wait for this to be merged and then make another PR based on the then-master.

thed636:
Hi, all!

> @thed636 I'll see what things I can make nicer using CMake 3.12. About changing to 3.14: I think the only improvement there is that the FindPostgreSQL.cmake file comes with imported targets, but I already work around it by including the file in the repository. If it doesn't have any other features besides that, I see no reason to require 3.14, it'd only make it more difficult for users.

Ok, if the upgrade to 3.14 is not necessary, well, it's good then. My most conservative part is happy about such a fact. :)  

> The real deal is providing a conanfile.py and perhaps a test package -- but given a proper CMakeLists.txt, as is provided by this PR, this is not a big deal either. I can try and incorporate them here, or just wait for this to be merged and then make another PR based on the then-master.

Yeah, PR based on the then-master is a really good approach. So you are the next. :)

@omartijn it looks like there is a problem with include path for examples as I can see from the CI.

omartijn:
@thed636 Yes, the include path broke when I merged commits from master into this repo. A test was added, which obviously used the old target_link_libraries syntax (because master doesn't yet have the `ozo` target) and git didn't recognize this as a conflict. I've updated this now.

omartijn:
Can anybody review this and let me know what, if anything, needs to be changed before this could be merged?

thed636:
Hi @omartijn, as for me, everything is ok. You need to squash all commits into one, add yourself into [AUTHORS](https://github.com/yandex/ozo/blob/master/AUTHORS), provide good commit message and rebase on current master, after that it could be merged. Thanks for your contribution!

**PS** Well, it seems like build on Mac OS is still broken.

omartijn:
@thed636 I have squashed, updated the commit message and added myself to the AUTHORS.

I don't know what to do about the Mac OS build. It is broken because we are using GNU-specific extensions. Perhaps we should just remove that? The problem is in include/ozo/query_builder.h on line 183.

thed636:
> I don't know what to do about the Mac OS build. It is broken because we are using GNU-specific extensions. Perhaps we should just remove that? The problem is in include/ozo/query_builder.h on line 183

Well, I doubt what we need to remove the useful code from the master which is passing CI success.

The [log](https://api.travis-ci.org/v3/job/556460278/log.txt) contains error:

```
[1m/Users/travis/build/yandex/ozo/include/ozo/query_builder.h:183:16: [0m[0;1;31merror: [0m[1mstring literal operator templates are a GNU extension [-Werror,-Wgnu-string-literal-operator-template][0m

constexpr auto operator "" _SQL() {

[0;1;32m               ^

[0m1 error generated.
```

Another interesting thing I that we have a warning:

```

[1m/Users/travis/build/yandex/ozo/include/ozo/query_builder.h:183:16: [0m[0;1;35mwarning: [0m[1mstring literal operator templates are a GNU extension [-Wgnu-string-literal-operator-template][0m

constexpr auto operator "" _SQL() {

[0;1;32m   
```

And it seems like somehow `-Wno-gnu-string-literal-operator-template` and `-Werror` flags set-up is missed.

include/ozo/query_builder.h
```diff
@@ -179,10 +179,14 @@ constexpr auto operator +(query_element<LhsValueT, LhsTagT>&& lhs, RhsValueT&& r
 
 namespace literals {
 
+#ifdef __GNUG__
+
```
thed636: Sorry, but this is wrong way. No source code should be modified on this PR. This PR should modify CMake files only. Enabling or disabling the library’s features is a completely different topic.

thed636 (changes requested):

CMakeLists.txt
```diff
@@ -4,6 +4,48 @@ project(ozo VERSION 0.0.1)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
+include(CheckCXXCompilerFlag)
+
+function (target_supported_compile_options target)
```
thed636: This is a good idea to have a function, but I think it would be nice to point certain flags for the certain compilers to be sure about flags set-up.

thed636 (changes requested):

thed636:
@omartijn in case if you have no ability to build on MacOS I can help you with it on the next week.

include/ozo/query_builder.h
omartijn: Well, the idea is of course to disable it for compilers that don't support this non-standard option. It won't change anything for supported compilers.

I've removed the guard, though I do think it should be added again because this prohibits people from using the library using different compilers.

omartijn:
I have fixed the compilation errors on mac os. The warning is now suppressed on compilers supporting the flag, this is also exported so that when you do `target_link_libraries` this option is automatically set. This prevents getting compilation warnings/errors from simply using ozo.

It of course still won't compile for compilers other than clang/gcc because of the non-standard features used. I'll open another PR for this.

CMakeLists.txt
thed636: Well, I meant it would be better to have an error if a compiler does not support an option. I would like to say what unsupported option should not be ignored silently because in that case, you can not have a guarantee that a given option is supported. E.g., in the case of a typo, there should be an error. So the silent strategy is not good in this case.

thed636 (changes requested):

CMakeLists.txt
```diff
@@ -1,20 +1,130 @@
...
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+
+include(CheckCXXCompilerFlag)
+
+function (target_supported_compile_options target)
+    set(options BEFORE)
+    set(arguments INTERFACE PUBLIC PRIVATE)
+    cmake_parse_arguments(OPTION "${options}" "" "${arguments}" ${ARGN})
+
+    foreach(compile_option ${OPTION_INTERFACE})
```
thed636: I think this code may be simplified if variable name would be generated like [this](https://stackoverflow.com/questions/32631168/cmake-interpret-string-as-variable)

benchmarks/CMakeLists.txt
```diff
@@ -1,33 +1,36 @@
...
+    target_compile_options(ozo_benchmark PRIVATE -Wno-ignored-optimization-argument)
+endif()
 
 add_executable(ozo_benchmark_performance performance.cpp)
-target_compile_features(ozo_benchmark_performance PRIVATE cxx_std_17)
-target_link_libraries(ozo_benchmark_performance ${LIBRARIES})
+target_link_libraries(ozo_benchmark_performance ozo)
+
+# enable a bunch of warnings and make them errors
+target_compile_options(ozo_benchmark_performance PRIVATE -Wall)
```
thed636: It is not necessary but would be nice to be able to make such declarations to reduce amount of code:

```
target_compile_options(ozo_benchmark PRIVATE -Wall -Wextra -Wsign-compare -pedantic -Werror)
```

or

```
set(GCC_CLANG_FLAGS "-Wall -Wextra -Wsign-compare -pedantic -Werror")

#...

target_compile_options(ozo_benchmark PRIVATE GCC_CLANG_FLAGS)
```

thed636 (changes requested):

CMakeLists.txt
omartijn: I don't really understand your point here, but I have removed the code now, in favor of hard-coding the flags.

benchmarks/CMakeLists.txt
omartijn: I have collapsed the compile options to reduce code amount.

CMakeLists.txt
omartijn: While I don't necessarily agree (I personally think it's nicer to do "feature detection" than "compiler detection"), I have changed it now.

thed636: Well, detection is a good idea, but if a feature is not supported, but needed - build should fail. In other cases - it is equivalent to the absence of a feature. Since different compilers require different features it is needed to bind feature set to the certain compiler(s). In other cases, requirements are just meaningless because no checks could be made.
To rely just on compiler errors is not a good strategy in terms of error source detection. Since sometimes it is too hard to understand relation about missed feature and compilation error.

CMakeLists.txt
thed636: The point is to do not have three almost identical `foreach` statement, instead of one which could be parametrized by the option's scope set.

thed636 (approved):

benchmarks/CMakeLists.txt
thed636: Congratulations, Martijn! Thanks for the help and good job!

PR #154, opened 2019-07-11
Disable the query builder operator on compilers not supporting it

omartijn:
This PR will disable the query builder operator on compilers other than gcc and clang. It won't compile anyway on different compilers.

### DIFF
--- a/include/ozo/query_builder.h
+++ b/include/ozo/query_builder.h
@@ -179,10 +179,14 @@ constexpr auto operator +(query_element<LhsValueT, LhsTagT>&& lhs, RhsValueT&& r
 
 namespace literals {
 
+#ifdef __GNUG__
+
 template <class CharT, CharT ... c>
 constexpr auto operator "" _SQL() {
     return make_query_builder(hana::make_tuple(make_query_text(hana::string<c ...>())));
 }
+
+#endif /* __GNUG__ */
 
 } // namespace literals
 } // namespace ozo


thed636:
Hi, Martijn!
Sorry, but this is still a wrong way. You can just try to disable the compiler feature and build the project with examples and tests. So, as a result, you'll see that this is so far from a solution to the problem. The direction is good, I mean usage of the GNU extension limits the supported compilers (e.g. we do not support VS for that reason). But the problem is much deeper.

omartijn:
> You can just try to disable the compiler feature

Which compiler feature are you talking about? We are using a feature which is not in the standard. This means ozo won't compile on many standards-compliant compilers, which is a shame. By only defining the operator on compilers that support it we can use ozo elsewhere too.

This PR has nothing to do with tests. The operator is defined in a header which is included _outside of the tests_ which means the whole library is unusable on compilers except gcc and clang.

thed636:
Short answer
=========

This PR does not solve the problem of using the library with compilers other than GCC and Clang for now. 

Long story
========

The same GNU extension is used with `boost::hana::string` via `BOOST_HANA_CONFIG_ENABLE_STRING_UDL`. So try to disable GNU literal operator extension and build the examples and tests of the project. You'll see the result. The library would not work without GNU extension of literal operator.

Next thing. In case you are trying to disable some of the library features for the compiler reason, it should be reflected in the documentation, tests, and CI. So just putting `#ifdef` and `#endif` is not enough for the PR success.

As I said the problem is much deeper and requires much more time and work.

omartijn:
Fair enough, I'll close this then.

PR #155, opened 2019-07-11
Implement support for timestamps

omartijn:
This PR will implement support for querying PG `timestamp without time zone` to an `std::chrono::system_clock::time_point`.

There are some things I don't exactly like, I'd be happy to hear any input on this:

* `std::chrono::system_clock::time_point` default constructor does not guarantee it is initialized to the UNIX epoch. The standard does suggest that _almost_ any implementation does this, and from c++20 it _is_ guaranteed. This means that the epoch I initialize could potentially be wrong. I don't know of any systems where this is the case, though.

* The magic number of days to add is also not the nicest way of doing it. Again, c++20 to the rescue, since that one has `std::chrono::years{ 30 }`, which we unfortunately can't use yet.

### DIFF
--- a/include/ozo/ext/std/time_point.h
+++ b/include/ozo/ext/std/time_point.h
@@ -1,0 +1,49 @@
+#pragma once
+
+#include <ozo/pg/definitions.h>
+#include <ozo/detail/epoch.h>
+#include <ozo/io/send.h>
+#include <ozo/io/recv.h>
+
+#include <chrono>
+#include <iostream>
+
+
+namespace ozo {
+
+
+/**
+ * @defgroup group-ext-std-chrono-system-clock-time-point std::chrono::system_clock::time_point
+ * @ingroup group-ext-std
+ * @brief [std::chrono::system_clock::time_point](https://en.cppreference.com/w/cpp/chrono/time_point) support
+ *
+ *@code
+#include <ozo/ext/std/time_point.h>
+ *@endcode
+ *
+ * `std::chrono::system_clock::time_point` is defined as a point in time.
+ */
+
+template <>
+struct send_impl<std::chrono::system_clock::time_point> {
+    template <typename M>
+    static ostream& apply(ostream& out, const oid_map_t<M>&, const std::chrono::system_clock::time_point& in) {
+        int64_t value = std::chrono::duration_cast<std::chrono::microseconds>(in - detail::epoch).count();
+        return impl::write(out, value);
+    }
+};
+
+template <>
+struct recv_impl<std::chrono::system_clock::time_point> {
+    template <typename M>
+    static istream& apply(istream& in, size_type, const oid_map_t<M>&, std::chrono::system_clock::time_point& out) {
+        int64_t value;
+        impl::read(in, value);
+        out = detail::epoch + std::chrono::microseconds{ value };
+        return in;
+    }
+};
+
+}
+
+OZO_PG_BIND_TYPE(std::chrono::system_clock::time_point, "timestamp")
--- a/include/ozo/detail/epoch.h
+++ b/include/ozo/detail/epoch.h
@@ -1,0 +1,12 @@
+#pragma once
+
+#include <chrono>
+
+namespace ozo::detail {
+
+// pg epoch is 2000-01-01 00:00:00, which is 30 years (10957 days)
+// after the unix epoch (1970-01-01 00:00:00, the default value for a time_point)
+// c++20: static constexpr const auto epoch = std::chrono::system_clock::time_point{} + std::chrono::years{ 30 };
+static constexpr const auto epoch = std::chrono::system_clock::time_point{} + std::chrono::hours{ 24 } * 10957;
+
+}
--- a/include/ozo/ext/std.h
+++ b/include/ozo/ext/std.h
@@ -19,3 +19,4 @@
 #include <ozo/ext/std/vector.h>
 #include <ozo/ext/std/weak_ptr.h>
 #include <ozo/ext/std/array.h>
+#include <ozo/ext/std/time_point.h>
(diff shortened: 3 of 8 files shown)


thed636:
Unit tests are required, should be similar to:

- [deserialization](https://github.com/yandex/ozo/blob/master/tests/binary_deserialization.cpp#L738-L760)
- [serialization](https://github.com/yandex/ozo/blob/master/tests/binary_serialization.cpp#L176-L189)

Integration test would be nice too, may be similar to:
- [integration](https://github.com/yandex/ozo/blob/master/tests/integration/result_integration.cpp#L36-L45)

**PS** Thanks for the PR, Martijn

include/ozo/ext/std/time_point.h
```diff
@@ -0,0 +1,55 @@
...
+#include <ozo/io/send.h>
+#include <ozo/io/recv.h>
+
+#include <chrono>
+#include <iostream>
+
+namespace ozo::pg {
+
+// pg epoch is 2000-01-01 00:00:00, which is 30 years (10957 days)
+// after the unix epoch (the default value for a time_point)
```
thed636: 1. Such thing should be described in the documentation if it is public.
2. In case you want to use `epoch` as a constant is related to PostgreSQL type system it is better to place it in header like [this](https://github.com/yandex/ozo/blob/master/include/ozo/pg/types/uuid.h).

include/ozo/ext/std/time_point.h
```diff
@@ -0,0 +1,55 @@
...
+static constexpr const auto epoch = std::chrono::system_clock::time_point{} + std::chrono::hours{ 24 } * 10957;
```
thed636: It is better to define `ozo::days`, `ozo::weeks`, `ozo::months`, `ozo::years` in and then define the epoch via this definitions (which would be using references for C++20 standard library).

thed636 (changes requested):

include/ozo/ext/std/time_point.h
omartijn: Yes, this would indeed be the nicest way to do it. This is, however, tricky to do with `std::chrono` in c++ < 20. It's fairly easy to define the concept of a day, since we have `std::chrono::hours` and a day is 24 hours. When you get to months it's more tricky, since the number of days in a month is not fixed.

The current implementation of a duration in `std::chrono` is that it's represented in a number of ticks, e.g. the number of microseconds. This is incompatible with the concept of months or years, since you can't represent them in a fixed number of these ticks.

Doing this would basically mean I can't use `std::chrono` (at least not in c++17).

include/ozo/ext/std/time_point.h
omartijn: I have considered splitting off the definition of this epoch, but could not find a proper place for it. The file you link to is in a directory named `types`, which this is certainly not. If you have an opinion on where you want this define, I'll be happy to place it there.

include/ozo/ext/std/time_point.h
thed636: > Yes, this would indeed be the nicest way to do it. This is, however, tricky to do with std::chrono in c++ < 20. It's fairly easy to define the concept of a day, since we have std::chrono::hours and a day is 24 hours. When you get to months it's more tricky, since the number of days in a month is not fixed.

Yes, you right, this is a big pain even in ProstgreSQL time intervals :)

include/ozo/ext/std/time_point.h
thed636: Well, if you do not want to expose it, the best choice (until modules) is to use `ozo::detail` namespace.

omartijn: That makes sense. Users of the library should never need this information, by the time they see the timestamp it should have already been converted to a proper time_point.

I have therefore moved it to the detail namespace.

omartijn:
I have added some unit tests. I'll look at making an integration test later.

thed636:
> I have added some unit tests. I'll look at making an integration test later.

Thanks, Martijn. 

I'd ask you to add a header which should be very similar to [ozo/pg/types/text.h](https://github.com/yandex/ozo/blob/master/include/ozo/pg/types/text.h) to define default `timestamp` implementation of the library and then add it [here](https://github.com/yandex/ozo/blob/master/include/ozo/pg/types.h).

So, after integration tests will be implemented the PR could be merged into master.

omartijn:
@thed636 I have added an integration test and added an `ozo::pg::timestamp` alias to `std::chrono::system_clock::time_point` as requested. If you agree with all this, I can rebase on master and squash.

Interesting thing is that many of the integration tests fail, including the one I just added, because the result for the text comes back as oid 705 (which is "unknown" according to ozo/pg/definitions.h.

If I cast the text explicitly to text by adding ::text, the tests work as expected. Since I can't imagine that someone would write all these tests and then add them in the failing state, I think this may be due to different PostgreSQL versions. It might be worthwhile to add an explicit cast to all those tests since I don't believe it'd break it for anyone.

thed636 (dismissed):

thed636:
> @thed636 I have added an integration test and added an `ozo::pg::timestamp` alias to `std::chrono::system_clock::time_point` as requested. If you agree with all this, I can rebase on master and squash.

Yes, everything is ok, so it would be really nice to do it. :)

> Interesting thing is that many of the integration tests fail, including the one I just added, because the result for the text comes back as oid 705 (which is "unknown" according to ozo/pg/definitions.h.
> 
> If I cast the text explicitly to text by adding ::text, the tests work as expected. Since I can't imagine that someone would write all these tests and then add them in the failing state, I think this may be due to different PostgreSQL versions. It might be worthwhile to add an explicit cast to all those tests since I don't believe it'd break it for anyone.

Well, the implicit type conversion is a dark corner of PostgreSQL. The best practice we found in production code is to specialize types explicitly. Unfortunately this approach is missed in the tests. If you have any corrections - PR would be helpful. We will pay attention to such things in future tests.

omartijn:
@thed636 done, rebased on master and squashed.

thed636 (approved):

thed636:
👍

PR #157, opened 2019-07-23
[WIP] Add conan support

JonasProgrammer:
Hi,

this PR adds support for packaging ozo with the [conan package manager](https://conan.io). It provides a package recipe that relies on the distribution generated by `cmake install` as well as a test package for that.

Furthermore, it uses the latest libpq and boost versions by default, though this can be overwritten by package consumers.

I updated `resource_pool` to the latest upstream version, so we can use conan and target based CMake for that as well.

Some notes: I had to make changes to the freshly accepted CMake PR. As it turns out, the installed package targets used the wrong include path, resulting in `ozo/` being stripped from headers for client code. The provided config also did not auto-require its dependencies, which I fixed.
For this however, I had to vendor in the FindPostgreSQL.cmake for the config step, as the one provided with CMake 3.12 did export `PostgreSQL::PostgreSQL`.
I also made the standalone package use `yandex::` as a target namespace. While this is not strictly required, it's what everybody else does. In theory this does break client code using the standalone package, but as it was merged only yesterday and had the include problems I doubt this really affects anyone.

I also had to make some ugly workarounds in travis. `conan` is required for testing the purpose and the Dockerfile is on ignore, so I figured adding it there might not be the best idea. Having it run on a `xenial` machine was not possible, as Hana complains about gcc being outdated. `bionic` then breaks the order for apt-addons and install, resulting in `pip3` not being present. I fixed this with an ugly workaround (that does not affect the other machines), but perhaps you have a better idea.

The PR is furthermore marked **WIP**. I would like to know what's yandex's/your standpoint regarding providing an official conan repo. For public OSS repos, bintray is free of charge. If this is desired, the travis script could be expanded to automatically building and deploying packages (given it's the master branch that is being built). As `resource_pool` doesn't have a public repo yet (AFAIK), it could also be uploaded there.
It would be a one-time setup but make the library much easier to use for users.

Sorry for the wall of text.

P.S.: Maybe not the PR you expected, but I hope it's useful nontheless

### DIFF
--- a/include/ozo/result.h
+++ b/include/ozo/result.h
@@ -10,13 +10,14 @@ namespace ozo {
 
 /**
  * @brief Database request result value proxy
- * @ingroup group-requests-types
  *
  * Provides access to values of a database request result. The library is designed
  * to do not obligate a user to have a deal with raw and untyped results representation.
  * But sometimes it is really needed to have an access to raw representation of the
  * result. E.g. for reducing memory consumption or performance reason. For that case
  * the library provides this mechanism.
+ *
+ * @ingroup group-requests-types
  */
 #ifndef OZO_DOCUMENTATION
 template <typename Result>
@@ -32,7 +33,7 @@ public:
     value(const coordinates& v) : v_(v) {}
 
     /**
-     * @brief Value type OID
+     * Get value type OID
      *
      * @return oid_t --- type OID
      */
@@ -41,7 +42,8 @@ public:
     }
 
     /**
-     * @brief Indicates text format representation for the value
+     * Determine whether the value is in binary format. Should be always false for current
+     * implementation.
      *
      * @return `true` --- value in text format
      * @return `false` --- value in binary format
@@ -51,7 +53,8 @@ public:
     }
 
     /**
-     * @brief Indicates binary format representation for the value
+     * Determine whether the value is in binary format. Should be always true for current
+     * implementation.
      *
      * @return `true` --- value in binary format
      * @return `false` --- value in text format
@@ -61,7 +64,7 @@ public:
     }
 
     /**
-     * @brief Value raw data buffer access
+     * Get value raw data buffer access.
      *
      * @return const char* --- pointer to raw data buffer
      * @sa ozo::value::size()
@@ -71,7 +74,7 @@ public:
     }
 
     /**
-     * @brief Value raw data buffer size
+     * Get value raw data buffer size.
      *
      * @return std::size_t --- size of raw data buffer in bytes.
      * @sa ozo::value::data()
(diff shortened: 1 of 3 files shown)


thed636:
Hi, Jonas!

Thanks for the contribution and such a good description! Could you please to make a separate PR for the `Update resource_pool` commit, I'd like to merge it into the master. The Conan repo is good, but this part needs more time to get the company position (if any) about Conan repo, and it looks like that the library needs versioning and releasing strategy which is honestly not even near to be ready. This is the right direction but takes some time.

JonasProgrammer:
I have done so, as requested.

If gathering an opinion regarding conan is a a time-consuming process, I could still simply remove the TODO-part regarding automated building and deployment of the conan package. It would have just been a nice-to-have.

Still, I think having a recipe in the official repo at least, would be useful and not hurt anyone. All changes made here should be 100% backward compatible, except for putting the installed package under `yandex::` (which I doubt affects anyone right now, as the installed package has a broken include path).
Note that this is just an alias to the `ozo` target, therefore it is still possible to link against it from tests or vendored builds using `add_subdirectory`, just not when using the standalone `find_package`.

Regarding the versioning, you could just do semantic versioning and bump the minor for every commit on master, as long as the major is 0 (IIRC breaking changes are fine for 0.x.y).

I wouldn't consider this a requirement for just adding the package recipe though. While would certainly be nicer with a versioning concept, it is possible to overwrite the package version on creation/deployment. It would be feasible, for example, to just append the commit ID to the dummy version. At least that's what I would do when uploading this to my private repo and then simply use the extended version when depending on ozo.

JonasProgrammer:
Hi Sergei,

do you have any news on this?

If conan is out of question for you, that's fine, but [this](https://github.com/yandex/ozo/blob/3360f13f9dddc247a7a8e04b4318c73c370e7c61/CMakeLists.txt#L37) is still problematic when using the library with `find_package` (out-of-tree from an install location) and [the lack of](https://github.com/yandex/ozo/pull/157/files#diff-a52b3843294abce5dd60d16cb15872c5) `find_dependency` forces you to manually specify these.

I could make a separate PR, if you wish. On the other hand, I still think having the recipe at least (I can remove `test_package/` and the changes to `.travis.yml` if they are deemed too intrusive) would not hurt anyone, but greatly improve the user experience, if someone is already using conan -- even without automated builds.

thed636:
Hi Jonas!

Regarding `find_dependency` - yes, the separate PR would be beneficial. Thanks for pointing on this problem. Since we use our build system inside the company, such attention from you is significant for us.

Regarding Conan repo. Yes, I'm still interested in this, however, now we are integrating OZO in some of the internal projects to replace old PostgreSQL client. Until this process ends, I do not want to produce any version since I expect some breaking changes in the customization points and, unlikely but possible, API. This process should be completed in this quarter. So I plan to return to this problem in December with the API stable version. And I'd be happy about your help in this process.

PR #159, opened 2019-07-25
Remove connection bound timer

thed636:
The connection-bound timer was a big architectural mistake.
It is very unsafe to rely on a shared timer while performing an operation. The operation-bound timer is a safe solution.
This PR uses deadline_handler to provide all the necessary time constraints instead of connection-bound-timer mechanisms.


elsid:
> It is very unsafe to rely on a shared timer while performing an operation.

Why is it unsafe? I can think about shared access that two parallel operations trying to use same timer for different timeouts.

thed636:
> > It is very unsafe to rely on a shared timer while performing an operation.
> 
> Why is it unsafe? I can think about shared access that two parallel operations trying to use same timer for different timeouts.

Well, the main problem is - an operation has no control of connection-bound timer state. I hit this problem at first attempt of `cancel()` implementation. I used a connection-bound timer for `cancel()` time-out. It was a mistake of course because it led to shared use of a timer between maybe-canceled operation and `cancel()` itself. So in the current implementation, I made a safer mechanism which allows a binding timer to an operation scope. An operation-bound timer is fully controlled by an operation scope and no such mistake could be made. And more - no more timer requirement for the `Connection` implementation.

include/ozo/impl/async_connect.h
```diff
@@ -104,16 +101,18 @@ struct async_connect_op {
     }
 
     void done(error_code ec = error_code {}) {
-        std::move(get_handler(context))(std::move(ec), std::move(get_connection(context)));
+        get_handler(context)(std::move(ec), std::move(get_connection(context)));
     }
 
-    using executor_type = std::decay_t<decltype(asio::get_associated_executor(get_handler(context)))>;
+    using handler_type = std::decay_t<decltype(get_handler(context))>;
+
```
serenityFX: ?

serenityFX (approved):

PR #167, opened 2019-08-29
Extend failover strategy with initiate_next_try()

thed636:
This mechanism supports static polymorphism of tries depended on error conditions.
Unlike `ozo::failover::get_next_try()` it allows to use different types of
try-objects for different error conditions.

Function implementation should call initiator in case of operation with given error
condition may be recovered by try-object is given to initiator. Or should not call
initiator in case of an unrecoverable error condition.

### DIFF
--- a/include/ozo/failover/role_based.h
+++ b/include/ozo/failover/role_based.h
@@ -11,6 +11,7 @@
 #include <boost/hana/size.hpp>
 #include <boost/hana/minus.hpp>
 #include <boost/hana/not_equal.hpp>
+#include <boost/hana/greater.hpp>
 
 /**
  * @defgroup group-failover-role_based Role-Based Execution & Fallback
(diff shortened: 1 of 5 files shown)


include/ozo/failover/role_based.h
```diff
@@ -515,26 +516,34 @@ class role_based_try {
      * @return `std::nullopt` --- retry is not possible due to `ec` value or tries remain
      *                            count.
      */
-    template <typename Connection>
-    auto get_next_try([[maybe_unused]] ozo::error_code ec, Connection&& conn) const {
```
serenityFX, line 519: Why Connection && -> &?

include/ozo/failover/role_based.h
```diff
@@ -515,26 +516,34 @@ class role_based_try {
...
+    template <typename Connection, typename Initiator>
+    void initiate_next_try([[maybe_unused]] ozo::error_code ec, Connection& conn, [[maybe_unused]] Initiator init) const {
```
serenityFX: Why not Initiator&&?

include/ozo/failover/role_based.h
```diff
@@ -591,14 +590,20 @@ class role_based_strategy : public options_factory_base<role_based_strategy<Opti
 
 /**
  * Try to perform an operation on first role with fallback to next items of
- * specified sequence in case of recoverable error.
+ * specified sequence which should recover an error.
+ * E.g., in case of sequence consists of <i>role1</i>, <i>role2</i>, <i>role3</i> and <i>role4</i>.
+ * If <i>role1</i> cause an error which could be recovered with <i>role3</i> and can not be
```
serenityFX, line 595: don't we want to add example to this case?

include/ozo/failover/role_based.h
thed636: Since the strategy does not transfer the ownership of a Connection object, there is no need to have a universal reference in the signature IMO.

include/ozo/failover/role_based.h
thed636: **Done**

include/ozo/failover/role_based.h
thed636: Yes we do, but I'm not ready to provide good example right now :(

serenityFX (approved):

PR #168, opened 2019-08-31
Allow customization of transaction initiation statement

JonasProgrammer:
- model postgres transaction options in code
- add options parameter to long-form of begin_op::operator()
- add .with_options()-helper to begin_op to control the implicit options

All credit for the interface ideas goes to @thed636

### DIFF
--- a/include/ozo/asio.h
+++ b/include/ozo/asio.h
@@ -201,7 +201,7 @@ constexpr execute_op<impl::initiate_async_execute> execute;
  * @endcode
  * @ingroup group-core-types
  */
-template <template<typename...> typename Operation, typename Initiator>
+template <typename Operation, typename Initiator>
 struct base_async_operation {
     using initiator_type = Initiator;
     using base = base_async_operation;
@@ -211,14 +211,10 @@ struct base_async_operation {
 
     constexpr initiator_type get_initiator() const { return initiator_;}
 
-    template <typename OtherInitiator>
-    constexpr static auto rebind_initiator(const OtherInitiator& other) {
-        return Operation<OtherInitiator>{other};
-    }
-
     template <typename InitiatorFactory>
     constexpr auto operator[] (InitiatorFactory&& f) const {
-        return rebind_initiator(construct_initiator(std::forward<InitiatorFactory>(f), *this));
+        const auto& op = static_cast<const Operation&>(*this);
+        return op.rebind_initiator(construct_initiator(std::forward<InitiatorFactory>(f), *this));
     }
 };
 
--- a/include/ozo/detail/begin_statement_builder.h
+++ b/include/ozo/detail/begin_statement_builder.h
@@ -1,0 +1,76 @@
+#pragma once
+
+#include <boost/hana/find.hpp>
+#include <boost/hana/filter.hpp>
+#include <boost/hana/fold.hpp>
+#include <boost/hana/not_equal.hpp>
+#include <boost/hana/optional.hpp>
+#include <boost/hana/plus.hpp>
+#include <boost/hana/string.hpp>
+#include <boost/hana/tuple.hpp>
+#include <ozo/core/none.h>
+#include <ozo/core/options.h>
+#include <ozo/query_builder.h>
+#include <ozo/transaction_options.h>
+#include <type_traits>
+
+namespace ozo::detail {
+
+namespace hana = boost::hana;
+
+struct begin_statement_builder {
+    using opt = transaction_options;
+
+    template <typename Level>
+    constexpr static auto to_string(decltype(opt::isolation_level), Level level) {
+        [[maybe_unused]] const auto prefix = BOOST_HANA_STRING(" ISOLATION LEVEL ");
+        const auto sql_levels = hana::make_map(
+            hana::make_pair(isolation_level::serializable, BOOST_HANA_STRING("SERIALIZABLE")),
+            hana::make_pair(isolation_level::repeatable_read, BOOST_HANA_STRING("REPEATABLE READ")),
+            hana::make_pair(isolation_level::read_committed, BOOST_HANA_STRING("READ COMMITTED")),
+            hana::make_pair(isolation_level::read_uncommitted, BOOST_HANA_STRING("READ UNCOMMITTED"))
+        );
+
+        if constexpr (const auto sql = hana::find(sql_levels, level); sql != hana::nothing) {
+            return prefix + *sql;
+        } else {
+            static_assert(std::is_void_v<Level>, "unexpected transaction level");
+        }
+    }
+
+    template <typename Mode>
+    constexpr static auto to_string(decltype(opt::mode), Mode mode) {
+        const auto sql_modes = hana::make_map(
+            hana::make_pair(transaction_mode::read_write, BOOST_HANA_STRING(" READ WRITE")),
+            hana::make_pair(transaction_mode::read_only, BOOST_HANA_STRING(" READ ONLY"))
+        );
+
+        if constexpr (const auto sql = hana::find(sql_modes, mode); sql != hana::nothing) {
+            return *sql;
+        } else {
+            static_assert(std::is_void_v<Mode>, "unknown transaction mode");
+        }
+    }
+
+    template <typename Deferrable>
+    constexpr static auto to_string(decltype(opt::deferrability), Deferrable) {
+        if constexpr (Deferrable::value) {
+            return BOOST_HANA_STRING(" DEFERRABLE");
+        } else {
+            return BOOST_HANA_STRING(" NOT DEFERRABLE");
+        }
+    }
+
+    template <typename Options>
+    constexpr static auto build(Options const& options) {
+        constexpr auto supported_options = hana::make_tuple(opt::isolation_level, opt::mode, opt::deferrability);
+        constexpr auto query_prefix = BOOST_HANA_STRING("BEGIN");
+
+        const auto real_options = hana::filter(hana::filter(supported_options, hana::partial(hana::contains, options)),
+                                               ([&](const auto& v) { return std::negation<is_none<std::decay_t<decltype(options[v])>>>{}; }));
+        const auto strings = hana::transform(real_options, [&](const auto& v) { return to_string(v, options[v]); });
+        return make_query(hana::fold(strings, query_prefix, hana::plus));
+    }
+};
+
+} // ozo::detail
--- a/include/ozo/transaction.h
+++ b/include/ozo/transaction.h
@@ -1,22 +1,28 @@
 #pragma once
 
+#include <ozo/core/options.h>
+#include <ozo/detail/begin_statement_builder.h>
 #include <ozo/impl/async_start_transaction.h>
 #include <ozo/impl/async_end_transaction.h>
+#include <ozo/transaction_status.h>
 
 namespace ozo {
 
-template <typename Initiator>
-struct begin_op : base_async_operation <begin_op, Initiator> {
+template <typename Initiator, typename Options = decltype(make_options())>
+struct begin_op : base_async_operation <begin_op<Initiator, Options>, Initiator> {
     using base = typename begin_op::base;
-    using base::base;
+    Options options_;
+
+    constexpr explicit begin_op(Initiator initiator = {}, Options options = {}) : base(initiator), options_(options) {}
+
     template <typename T, typename TimeConstraint, typename CompletionToken>
     auto operator() (T&& provider, TimeConstraint t, CompletionToken&& token) const {
         static_assert(ConnectionProvider<T>, "provider should be a ConnectionProvider");
         static_assert(ozo::TimeConstraint<TimeConstraint>, "should model TimeConstraint concept");
         using namespace ozo::literals;
-        return async_initiate<CompletionToken, handler_signature<impl::transaction<connection_type<T>>>>(
+        return async_initiate<CompletionToken, handler_signature<impl::transaction<connection_type<T>, Options>>>(
             get_operation_initiator(*this), token,
-            std::forward<T>(provider), "BEGIN"_SQL, t);
+            std::forward<T>(provider), options_, detail::begin_statement_builder::build(options_), t);
     }
 
     template <typename T, typename CompletionToken>
@@ -27,13 +33,23 @@ struct begin_op : base_async_operation <begin_op, Initiator> {
             std::forward<CompletionToken>(token)
         );
     }
+
+    template <typename OtherOptions>
+    auto with_transaction_options(const OtherOptions& options) const {
+        return begin_op<Initiator, OtherOptions>{get_operation_initiator(*this), options};
+    }
+
+    template <typename OtherInitiator>
+    constexpr auto rebind_initiator(const OtherInitiator& other) const {
+        return begin_op<OtherInitiator, Options>{other, options_};
+    }
 };
 
 constexpr begin_op<impl::initiate_async_start_transaction> begin;
 
 struct commit_op {
-    template <typename T, typename TimeConstraint, typename CompletionToken>
-    auto operator() (impl::transaction<T>&& transaction, TimeConstraint t, CompletionToken&& token) const {
+    template <typename T, typename Options, typename TimeConstraint, typename CompletionToken>
+    auto operator() (impl::transaction<T, Options>&& transaction, TimeConstraint t, CompletionToken&& token) const {
         static_assert(ozo::TimeConstraint<TimeConstraint>, "should model TimeConstraint concept");
         using namespace ozo::literals;
         return async_initiate<CompletionToken, handler_signature<T>>(
@@ -41,8 +57,8 @@ struct commit_op {
             std::move(transaction), "COMMIT"_SQL, t);
     }
 
-    template <typename T, typename CompletionToken>
-    auto operator() (impl::transaction<T>&& transaction, CompletionToken&& token) const {
+    template <typename... Ts, typename CompletionToken>
+    auto operator() (impl::transaction<Ts...>&& transaction, CompletionToken&& token) const {
         return (*this)(
             std::move(transaction),
             none,
@@ -54,8 +70,8 @@ struct commit_op {
 constexpr commit_op commit;
 
 struct rollback_op {
-    template <typename T, typename TimeConstraint, typename CompletionToken>
-    auto operator() (impl::transaction<T>&& transaction, TimeConstraint t, CompletionToken&& token) const {
+    template <typename T, typename Options, typename TimeConstraint, typename CompletionToken>
+    auto operator() (impl::transaction<T, Options>&& transaction, TimeConstraint t, CompletionToken&& token) const {
         static_assert(ozo::TimeConstraint<TimeConstraint>, "should model TimeConstraint concept");
         using namespace ozo::literals;
         return async_initiate<CompletionToken, handler_signature<T>>(
@@ -63,8 +79,8 @@ struct rollback_op {
             std::move(transaction), "ROLLBACK"_SQL, t);
     }
 
-    template <typename T, typename CompletionToken>
-    auto operator() (impl::transaction<T>&& transaction, CompletionToken&& token) const {
+    template <typename... Ts, typename CompletionToken>
+    auto operator() (impl::transaction<Ts...>&& transaction, CompletionToken&& token) const {
         return (*this)(
             std::move(transaction),
             none,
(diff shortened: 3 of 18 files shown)


JonasProgrammer:
A note about `.with_options()`: I'm not really satisfied with how I implemented that, but I have no idea how to do it better. Adding the default options as a member to `begin_op` itself is kinda troublesome, as implementing a constructor to set it would require relying on implementation details of `basic_async_operation`.
The helper currently holds a copy of the `begin_op` instance, which I don't really like either, but references seemed to risky. Conditionally using a reference or a copy depending on whether `.with_options()` is being called on an lvalue or rvalue would be an option, but still there would be the problem of `begin_op base{}; return base.with_options(foo);` and so on.

If you have any idea on how to make this more efficient without compromising its usage safety, I'm open for improvements.

include/ozo/transaction_options.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+struct isolation_level {
+    constexpr static struct serializable_t {
+    } serializable{};
+    constexpr static struct repeatable_read_t {
+    } repeatable_read{};
+    constexpr static struct read_committed_t {
+    } read_committed{};
+    constexpr static struct read_uncommitted_t {
+    } read_uncommitted{};
```
thed636: It is better to use `boost::hana::type` because in this case, you will get Comparable constants

```c++
struct isolation_level {
    constexpr static hana::type<class serializable_tag> serializable{};
    constexpr static hana::type<class repeatable_read_tag> repeatable_read{};
    constexpr static hana::type<class read_committed_tag> read_committed{};
    constexpr static hana::type<class read_uncommitted_tag> read_uncommitted{};
};
```

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,85 @@
...
+#include <ozo/core/none.h>
+#include <ozo/core/options.h>
+#include <ozo/query_builder.h>
+#include <ozo/transaction_options.h>
+#include <type_traits>
+
+namespace ozo::detail {
+
+struct begin_statement_builder {
+    constexpr static auto empty_sql() {
```
thed636: Maybe the constant is better?

```c++
static constexpr auto empty_sql = hana::string_c<>;
```

Since there is no parameter in your query, it is ok to use `hana::string` for the text part and then use `ozo::make_query(text)` to create an open transaction query.

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,85 @@
...
+        using namespace ozo::literals;
+        return ""_SQL;
+    }
+
+    template <typename Level, typename = std::enable_if_t<is_isolation_level_v<Level>>>
+    constexpr static auto level_part(Level) {
+        using namespace ozo::literals;
+
+        auto prefix = " ISOLATION LEVEL "_SQL;
+        if constexpr (std::is_same_v<Level, isolation_level::serializable_t>) {
```
thed636: So if `isolation_level` entries are defined as `hana::type_c`, you can use `operator ==` like this:

```c++
template <typename Level>
constexpr static auto level_part(Level level) {
    const auto prefix = BOOST_HANA_STRING(" ISOLATION LEVEL ");
    if constexpr (decltype(level==none)::value) {
        return empty_sql;
    } else if constexpr (decltype(level == isolation_level::serializable)::value) {
        return prefix + BOOST_HANA_STRING("SERIALIZABLE");
   //...
    } else {
        // For the unknown type we will generate an error message
        static_assert(std::is_void_v<Level>, "unknown isolation level");
    }
}
```

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,85 @@
...
+            return std::move(prefix) + "READ UNCOMMITTED"_SQL;
+        }
+    }
+
+    constexpr static auto level_part(none_t) {
+        return empty_sql();
+    }
+
+    template <typename Mode, typename =  std::enable_if_t<is_transaction_mode_v<Mode>>>
+    constexpr static auto mode_part(Mode) {
```
thed636: The same thing here:

```c++
template <typename Mode>
constexpr static auto mode_part(Mode mode) {
    if constexpr (decltype(mode == none)::value) {
        return empty_sql;
    } else if constexpr (decltype(mode == transaction_mode::read_write)::value) {
       return BOOST_HANA_STRING(" READ WRITE");
    //...
    } else {
        static_assert(std::is_void_v<Mode>, "unknown transaction mode");
    }
}
```

include/ozo/transaction.h
```diff
@@ -54,17 +84,17 @@ struct commit_op {
...
         using namespace ozo::literals;
         return async_initiate<CompletionToken, handler_signature<T>>(
             impl::initiate_async_end_transaction{}, token,
             std::move(transaction), "ROLLBACK"_SQL, t);
     }
 
-    template <typename T, typename CompletionToken>
-    auto operator() (impl::transaction<T>&& transaction, CompletionToken&& token) const {
+    template <typename T, typename Options, typename CompletionToken>
+    auto operator() (impl::transaction<T, Options>&& transaction, CompletionToken&& token) const {
```
thed636: I think it is ok to use variadic params for the transaction, like:

```c++
template <typename CompletionToken, typename ...Ts>
auto operator() (impl::transaction<Ts...>&& transaction, CompletionToken&& token) const
```

include/ozo/transaction_options.h
```diff
@@ -0,0 +1,76 @@
...
+};
+
+template <typename T>
+constexpr bool is_isolation_level_v = std::is_same_v<T, isolation_level::serializable_t> ||
```
thed636: And you do not need such things like this.

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,85 @@
...
+    template <typename Deferrable, typename = std::enable_if_t<is_deferrable_mode_v<Deferrable>>>
+    constexpr static auto deferrable_part(Deferrable deferrable) {
+        using namespace ozo::literals;
+
+        auto deferrable_sql = " DEFERRABLE"_SQL;
+        if constexpr (!deferrable()) {
+            return " NOT"_SQL + std::move(deferrable_sql);
+        } else {
+            return deferrable_sql;
+        }
```
thed636, line 61: In the same way there maybe something like this:

```c++
template <typename Deferrable>
constexpr static auto deferrable_part(Deferrable deferrable) {
    if constexpr (decltype(level == none)::value) {
        return empty_sql;
    } else if constexpr (Deferrable::value) {
       return deferrable_sql;
    } else {
        return BOOST_HANA_STRING(" NOT") + deferrable_sql;
    }
}
```

thed636 (changes requested):
It looks like the idea of common `base_async_operation::rebind_initiator` is limiting us here.

Let's try something like this:

```c++
template <typename Operation, typename Initiator>
struct base_async_operation {
    using initiator_type = Initiator;
    using base = base_async_operation;
    initiator_type initiator_;

    constexpr base_async_operation(const Initiator& initiator = Initiator{}) : initiator_(initiator) {}

    constexpr initiator_type get_initiator() const { return initiator_;}

    template <typename InitiatorFactory>
    constexpr auto operator[] (InitiatorFactory&& f) const {
        const Operation& op = static_cast<const Operation&>(*this);
        return op.rebind_initiator(construct_initiator(std::forward<InitiatorFactory>(f), op));
    }
};
```

And each operation should define `rebind_initiator`. In this case, options may become a member of `begin_op`.

```c++
template <typename Initiator, typename Options>
struct begin_op : base_async_operation <begin_op, Initiator> {
    using base = typename begin_op::base;
    Options options_;
    
    begin_op(Initiator initiator, Options options) : base(initiator), options_(options) {}
    //...
    template <typename OtherInitiator>
    constexpr auto rebind_initiator(const OtherInitiator& other) const {
        return begin_op <OtherInitiator, Options>{other, options_};
    }
};
```

include/ozo/detail/begin_statement_builder.h
JonasProgrammer: I did it in this way, but this actually fails for the case when level is indeed none, as the comparison operator for `none_t` is called, thus resulting in bool being decltyped.

I fixed this and the other stuff above accordingly, but for the none-case I simply sticked to calling `is_none<T>::value` for `Deferrable`, `Mode` and `Level`.

JonasProgrammer:
Sorry for coming back to you only now, I did not have time earlier.

I think I have addressed all of the issues, but I had to make the aforementioned changes to the detection of ozo::none, as it's comparison operator broke the 'hanaized' way of doing this. While I could have circumvented this by adding a `hana::type` for empty transaction options, I'm not really sure this would improve things -- while it would allow the use of the hana syntax, it would introduce yet another way of saying 'nothing', which is exactly the point of `none(_t)` (at least IMO).
Also I had to make the `Operation` argument a concrete type, as I did not find any good way to get the original template template arguments in `operator[]` of `basic_async_operation`; if you have a better idea on how to go about this, let me now.

Also thank you for the patience with me; I lack experience in both Boost Hana (though, the more of it I see, the more I want to really get into it) and meta programming in general.

tests/integration/transaction_integration.cpp
```diff
@@ -54,4 +54,148 @@ TEST(transaction, create_schema_in_transaction_and_rollback_then_table_should_no
     io.run();
 }
 
+TEST(transaction, transaction_level_options_should_work) {
+    // start and commit a transaction with every possible option (mostly to ensure there's no typos in the generated SQL parts)
```
thed636: Well, in this case, it is better to rename to

```TEST(transaction, transaction_level_options_should_not_cause_sql_syntax_error)```

And the same things for the others.

File: include/ozo/transaction.h [hunk 9]
```diff
@@ -1,22 +1,38 @@
...
     using base = typename begin_op::base;
-    using base::base;
-    template <typename T, typename TimeConstraint, typename CompletionToken>
-    auto operator() (T&& provider, TimeConstraint t, CompletionToken&& token) const {
+    Options options_;
+
+    constexpr explicit begin_op(Initiator initiator = {}, Options options = {}) : base(initiator), options_(options) {}
+
+    template <typename T, typename TimeConstraint, typename ConcreteOptions, typename CompletionToken>
+    auto operator() (T&& provider, TimeConstraint t, ConcreteOptions&& options, CompletionToken&& token) const {
```
thed636: Well, I think there should be only one method to set-up options to do not provide a misleading user interface. I thought about an additional argument to `operator ()`, but such interface would have a combinatorial explosion of functions in case of trying to keep it consistent. So that's why `with_options()` was the proposal to the interface. Is there any use-case issue or inconvenience to keep both interface solutions?

include/ozo/transaction.h
```diff
@@ -75,4 +101,52 @@ struct rollback_op {
 
 constexpr rollback_op rollback;
 
+/**
+ * @ingroup group-connection-functions
+ * @brief Retrieve's a transactions isolation level
+ *
+ * Retrieve's a transaction's isolation level (that is, an instance of one of the child types of ozo::isolation_level) or an instance of ozo::none_t, if no
```
thed636: There is no need to repeat `@brief` section. Also, you can use `scripts/build.sh docs` and look at the result in your browser by opening `build/docs/html/index.html`.

include/ozo/transaction.h
```diff
@@ -75,4 +101,52 @@ struct rollback_op {
...
+ * such thing could be found.
+ *
+ * @tparam T transaction type
+ * @param transaction transaction to get the isolation level from
+ * @return isolation level or none
+ */
+template<typename T>
+constexpr auto get_transaction_isolation_level(T&& transaction) {
```
thed636: Since you are using a template parameter `T` instead of `impl::transaction<Ts...>` it is better to make a static assertion which emulates concept assertion about `T`, to provide a better error message. Or just use `const impl::transaction<Ts...>&` as an argument type.

include/ozo/transaction_options.h
```diff
@@ -0,0 +1,54 @@
...
+namespace ozo {
+
+namespace hana = boost::hana;
+
+/**
+ * @ingroup group-connection-types
+ * @brief 'type enum' for transaction isolation levels supported by PostgreSQL
+ */
+struct isolation_level {
+    constexpr static hana::type<class serializable_tag> serializable{};
```
thed636: It is better to make a short comment about each element like

```c++
constexpr static hana::type<class serializable_tag> serializable{};` //!< Serializable isolation level
```

In other cases no item would be displayed in the documentation. Try to generate it and read it.

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,71 @@
...
+#include <type_traits>
+
+namespace ozo::detail {
+
+struct begin_statement_builder {
+    static constexpr auto empty_sql = hana::string_c<>;
+
+    template <typename Level>
+    constexpr static auto level_part(Level level) {
+        auto prefix = BOOST_HANA_STRING(" ISOLATION LEVEL ");
```
thed636: [`[[maybe_unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute is better than `void` cast.

`auto` -> `const auto` to explicitly require no modification on this variable.

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,71 @@
...
+        } else {
+            static_assert(std::is_void_v<Mode>, "unknown transaction mode");
+        }
+    }
+
+    template <typename Deferrable>
+    constexpr static auto deferrable_part(Deferrable) {
+        if constexpr (is_none<std::decay_t<Deferrable>>::value) {
+            return empty_sql;
+        } else if constexpr (auto deferrable_sql = BOOST_HANA_STRING(" DEFERRABLE"); Deferrable::value) {
```
thed636: `auto` -> `const auto`

It's up to you, but IMO there are no needs to use is-statement `deferrable_sql` since it is a zero-byte value and fully compile-time thing.

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,71 @@
...
+    constexpr static auto build(Options const& options) {
+        auto const level = get_option(options, transaction_options::isolation_level, none);
+        auto const mode = get_option(options, transaction_options::mode, none);
+        auto const deferrability = get_option(options, transaction_options::deferrability, none);
+
+        return ozo::make_query(BOOST_HANA_STRING("BEGIN") +
+                               begin_statement_builder::level_part(level) +
+                               begin_statement_builder::mode_part(mode) +
+                               begin_statement_builder::deferrable_part(deferrability));
+    }
```
thed636, line 73: You can try some magic of boost::hana, which allows you to make more generic and linear code:

```c++
using opt = transaction_options;
constexpr auto supported_options = hana::make_tuple(opt::isolation_level, opt::mode, opt::deferrability);
constexpr auto query_prefix = BOOST_HANA_STRING("BEGIN");
//...
template <typename Options>
constexpr static auto build(Options const& options) {
    const auto real_options = hana::filter(supported_options, hana::partial(hana::contains, options));
    const auto strings = hana::transform(real_options, [&](auto& v) { return to_string(v, options[v]);});
    return hana::fold(strings, query_prefix, hana::plus);
}
```

In this case, you do not need to handle `ozo::none`, and can use `to_string` methods instead of `*_part` ones and dispatch it, for example, by the first argument.

include/ozo/detail/begin_statement_builder.h
thed636: See the comment below :)

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,71 @@
...
+        } else if constexpr (decltype(level == isolation_level::serializable)::value) {
+            return prefix + BOOST_HANA_STRING("SERIALIZABLE");
+        } else if constexpr (decltype(level == isolation_level::repeatable_read)::value) {
+            return prefix + BOOST_HANA_STRING("REPEATABLE READ");
+        } else if constexpr (decltype(level == isolation_level::read_committed)::value) {
+            return prefix + BOOST_HANA_STRING("READ COMMITTED");
+        } else if constexpr (decltype(level == isolation_level::read_uncommitted)::value) {
+            return prefix + BOOST_HANA_STRING("READ UNCOMMITTED");
+        } else {
+            static_assert(std::is_void_v<Level>, "transaction level");
```
thed636: It is better to provide a more user-friendly message, like "unexpected transaction level", or something like this, as you did it below. :)

thed636 (changes requested):
Hi, Jonas!

Thanks for the help! Hope you enjoy the project. I made some comments, and I think we have big progress here. Hana is a really good library for modern C++ metaprogramming which allows to do a lot of things in compile-time in a convenient manner.

include/ozo/detail/begin_statement_builder.h
JonasProgrammer: Yup, sorry, this was a quick&dirty fix because gcc and clang behave differently for -Wunused-variable when it comes to `if constexpr` paths.

include/ozo/detail/begin_statement_builder.h
JonasProgrammer: For some reason I misspell 'deferrable' 9/10 times on the first try, so this was more or less added out of spite. Removed for better readability.

include/ozo/transaction.h
JonasProgrammer: My bad, I was under the impression that @brief is used in the function list and the 'body' in the detailed method description (and as you correctly guessed, I never looked at it in its generated form).

include/ozo/transaction.h
JonasProgrammer: Well, this is a tough one, but IMHO both interfaces have their use cases. The parameter on `operator ()` is more in line with time constraints and so on.
However, the `with_options` helper is really useful, when you have to use the same isolation level (I need pretty much all of mine to be `SERIALIZABLE`, for example); doing this at connection level is not a good option, when using a shared conn pool. Having to type out the parameter everywhere on the other hand is extra work you can forget at times, when a simple `constexpr auto begin_serial = begin.with_options(...)` makes it clear from the 'function name' alone, that you use your required options, without having to skim through the arg list.

include/ozo/asio.h
```diff
@@ -211,14 +211,10 @@ struct base_async_operation {
 
     constexpr initiator_type get_initiator() const { return initiator_;}
 
-    template <typename OtherInitiator>
-    constexpr static auto rebind_initiator(const OtherInitiator& other) {
-        return Operation<OtherInitiator>{other};
-    }
-
-    template <typename InitiatorFactory>
+    template <typename InitiatorFactory, typename... OperationArgs>
```
thed636: It seems like `typename... OperationArgs` is redundant.

include/ozo/transaction.h
```diff
@@ -1,22 +1,38 @@
...
+#include <ozo/detail/begin_statement_builder.h>
 #include <ozo/impl/async_start_transaction.h>
 #include <ozo/impl/async_end_transaction.h>
+#include <ozo/transaction_status.h>
 
 namespace ozo {
 
-template <typename Initiator>
-struct begin_op : base_async_operation <begin_op, Initiator> {
+template <typename Initiator, typename StatementBuilder, typename Options = decltype(make_options())>
```
thed636: Does it really need to point `StatementBuilder` as a template parameter? Does it really need to be customized? I mean what is the customization point it provides?

include/ozo/transaction.h
thed636: Well, I have some arguments against the dedicated version of `operator()`:
1. Yet another version with options has an additional cost (even not so hight for now) for the compilation time.
2. An additional cost to support it.
3. User would ask - why the library does not provide `begin(Provider, Options, Token)` version when I'm using my own timer for the cancelation?

So IMO it is better to rename `with_options` (if it looks like an inconvenient name) to something more laconic than having an additional version of `operator()`.

include/ozo/transaction.h
```diff
@@ -75,4 +101,50 @@ struct rollback_op {
...
 
+/**
+ * @ingroup group-connection-functions
+ * @brief Retrieve's a transactions isolation level
+ *
+ * The isolation level can either be an instance of one of the child types of ozo::isolation_level or
+ * `ozo::none` if no level was explicitly set.
+ *
+ * @param transaction transaction to get the isolation level from
+ * @return isolation level or none
```
thed636: It is better to write something like this

```
@return ozo::isolation_level --- if the isolation level has been set
@return ozo::none --- if no isolation level has been specified for the transaction
```

In this case, necessary links should be generated by Doxygen.

include/ozo/transaction_options.h
```diff
@@ -0,0 +1,61 @@
...
+#include <ozo/core/options.h>
+#include <type_traits>
+
+namespace ozo {
+
+namespace hana = boost::hana;
+
+/**
+ * @ingroup group-connection-types
+ * @brief 'type enum' for transaction isolation levels supported by PostgreSQL
```
thed636, line 15: I'd like to ask you to provide for this and the other options a link for the related PostgreSQL documentation here, to allow a user easily navigate to it.

tests/detail/begin_statement_builder.cpp
```diff
@@ -0,0 +1,211 @@
...
+              get_text(begin_statement_builder::build(make_options())));
+
+    EXPECT_EQ(get_text(begin_statement_builder::build(make_options(transaction_options::deferrability = ozo::none,
+                                                                   transaction_options::mode = transaction_mode::read_only))),
+              get_text(begin_statement_builder::build(make_options(transaction_options::mode = transaction_mode::read_only))));
+}
+
+struct my_true_t { constexpr static bool value = true; } my_true;
+
+TEST(begin_statement_builder, should_allow_integral_constants_for_deferrability) {
```
thed636, line 189: I'm not sure about this test is necessary. You provide `deferrable` expectations in the previous test. I do not think that check for support custom true or false type is really needed.

include/ozo/detail/begin_statement_builder.h
```diff
@@ -0,0 +1,69 @@
...
+        [[maybe_unused]] const auto prefix = BOOST_HANA_STRING(" ISOLATION LEVEL ");
+        if constexpr (decltype(level == isolation_level::serializable)::value) {
+            return prefix + BOOST_HANA_STRING("SERIALIZABLE");
+        } else if constexpr (decltype(level == isolation_level::repeatable_read)::value) {
+            return prefix + BOOST_HANA_STRING("REPEATABLE READ");
+        } else if constexpr (decltype(level == isolation_level::read_committed)::value) {
+            return prefix + BOOST_HANA_STRING("READ COMMITTED");
+        } else if constexpr (decltype(level == isolation_level::read_uncommitted)::value) {
+            return prefix + BOOST_HANA_STRING("READ UNCOMMITTED");
+        } else {
```
thed636, line 36: It is not really mandatory, but since `isolation_level` members are `hana::type` you can try to use [hana::map](https://www.boost.org/doc/libs/1_67_0/libs/hana/doc/html/structboost_1_1hana_1_1map.html) here for dispatching. I think you could like the results. :)

tests/integration/transaction_integration.cpp
thed636: I think the comments should be removed as you made such a good name for the tests :)

thed636 (changes requested):
Hi, Jonas!

Sorry for the delay caused by C++Con attend. I made some comments on this version of PR. It much better now, but we still have to solve some problems. Thanks for your help and patient. :)

With best regards, 
Sergei.

include/ozo/transaction.h
JonasProgrammer: The original intention was to have it customizable so you could, in theory, roll your own statement builder, e.g. when a newer PostgreSQL version, extension or fork would add additional options for the begin statement.
That being said, this is indeed very unlikely, so I did remove it. If anyone ever needs it, it's easy enough for them to put it back in.

JonasProgrammer:
Hi Sergei,

first of all I hope CppCon was as interesting IRL as it look from the recordings I have watched so far. Also thanks again for the patience with me; I think I have addressed all points now. After all, I think you are right about `operator()` on `begin_op`, it just felt natural to add it because of how you handle timeouts.
I also used `hana::map` at the suggested places, but still kept an `if constexpr` to catch the invalid type case -- I am aware of `hana::at_key`, but IMHO having a static assert that clearly states the issue is better than having someone, who accidentally got some option values mixed up, walk through the error messages originating from the depths of `hana`.

EDIT: As for CI, I have no idea why thread-san broke. The very same commit [works fine for me](https://travis-ci.org/JonasProgrammer/ozo/builds/590532564).

thed636 (approved):

thed636:
Hi, Jonas!

Thanks for your contribution!

PR #173, opened 2019-10-02
Support PostgreSQL interval as std::microseconds

bmi-x:


tests/binary_serialization.cpp
```diff
@@ -198,4 +198,14 @@ TEST_F(send, with_std_chrono_time_point_should_store_as_microseconds) {
     }));
 }
 
+TEST_F(send, with_std_chrono_microseconds_should_store_as_days_and_microseconds) {
+    const auto microseconds = (((((((7LL * 12 + 8) * 30 + 9) * 24 + 10) * 60 + 11) * 60 + 12) * 1000 + 13) * 1000 + 14);
```
thed636: What are these calculations for?

tests/integration/result_integration.cpp
```diff
@@ -55,6 +55,16 @@ TEST(result, should_convert_into_tuple_time_point_and_text) {
     EXPECT_EQ(std::get<1>(r[0]), "2");
 }
 
+TEST(result, should_convert_into_tuple_microseconds) {
+    auto result = execute_query("SELECT '7 years 8 months 9 days 10 hours 11 minutes 12 seconds 13 milliseconds 14 microseconds'::interval");
+    auto oid_map = ozo::empty_oid_map();
+    ozo::rows_of<std::tuple<std::chrono::microseconds>> rows;
```
thed636: You do not need to point a tuple of microseconds - the purpose of shortcut `ozo::rows_of` is to omit tuple, see the [docs](https://yandex.github.io/ozo/group__group-requests-types.html#gade9a42efa5064524382c2673de242f39) for details.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,83 @@
...
+
+        using microseconds = std::chrono::duration<std::int64_t, std::micro>;
+        using days = std::chrono::duration<std::int64_t, std::ratio<24LL * std::chrono::hours::period::num>>;
+        using months = std::chrono::duration<std::int64_t, std::ratio<30LL * days::period::num>>;
+
+        detail::pg_interval interval;
+        read(in, interval);
+
+        auto duration_microseconds = microseconds(interval.microseconds) % days(1);
+        auto duration_days_total = (days(months(interval.months)) + days(interval.days) + std::chrono::duration_cast<days>(microseconds(interval.microseconds)));
```
thed636: It seems like you do not need `days(months(interval.months))` since the common type for `days` and `months` should be `days`.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,83 @@
...
+    }
+};
+
+template <>
+struct recv_impl<std::chrono::microseconds> {
+    template <typename OidMap>
+    static istream& apply(istream& in, size_type, const OidMap&, std::chrono::microseconds& out) {
+        static_assert(ozo::OidMap<OidMap>, "OidMap should model ozo::OidMap");
+
+        using microseconds = std::chrono::duration<std::int64_t, std::micro>;
```
thed636: It is not safe to declare your own `microseconds` since if `std::chrono::microseconds` has different representation type you may have undefined behavior. The best solution is static assert on the representation type traits.

include/ozo/ext/std/duration.h
thed636: IMO prefix `duration_` is redundant since this is a variables' type property. Too long variable names do not make the code understanding easer. I suggest something `days_total` and `us_surplus` as variable names.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,83 @@
...
+ * `std::chrono::microseconds` is mapped as `interval` PostgreSQL type.
+ */
+
+template <>
+struct send_impl<std::chrono::microseconds> {
+    template <typename OidMap>
+    static ostream& apply(ostream& out, const OidMap&, const std::chrono::microseconds& in) {
+        static_assert(ozo::OidMap<OidMap>, "OidMap should model ozo::OidMap");
+
+        using days = std::chrono::duration<std::int64_t, std::ratio<24LL * std::chrono::hours::period::num>>;
```
thed636: IMO it is better to use `int32_t` representation instead of `int64_t` representation and type downcast.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,83 @@
...
+
+        write(out, detail::pg_interval {(in % days(1)).count(), std::int32_t(in / days(1)), 0});
```
thed636: IMO `std::chrono::duration_cast<days>(in)` is safer than arithmetic operation with type downcast.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,83 @@
...
+
+template <>
+struct recv_impl<std::chrono::microseconds> {
+    template <typename OidMap>
+    static istream& apply(istream& in, size_type, const OidMap&, std::chrono::microseconds& out) {
+        static_assert(ozo::OidMap<OidMap>, "OidMap should model ozo::OidMap");
+
+        using microseconds = std::chrono::duration<std::int64_t, std::micro>;
+        using days = std::chrono::duration<std::int64_t, std::ratio<24LL * std::chrono::hours::period::num>>;
+        using months = std::chrono::duration<std::int64_t, std::ratio<30LL * days::period::num>>;
```
thed636: It seems like `std::int32_t` fits better for months.

tests/binary_deserialization.cpp
```diff
@@ -776,4 +776,78 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
     EXPECT_EQ(result, expected);
 }
 
+struct recv_microseconds_ok : recv, WithParamInterface<std::pair<std::vector<char>, std::chrono::microseconds>> {
+    static_assert(std::chrono::microseconds::min().count() == std::numeric_limits<int64_t>::min());
+    static_assert(std::chrono::microseconds::max().count() == std::numeric_limits<int64_t>::max());
```
thed636: What are these assertions for?

tests/binary_deserialization.cpp
```diff
@@ -776,4 +776,78 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+    make_param({0xFF, 0xFF, 0xFF, 0x83, 0xEF, 0x61, 0x60, 0x00, 0xDE, 0xE2, 0xE5, 0x1D, 0x00, 0xE4, 0x45, 0xF7}, -9223372036854775808ULL),  // 14'960'119 m + -555'555'555 d + -532'854'775'808 us (minimum)
+
+    make_param({0x00, 0x00, 0x00, 0x03, 0x5D, 0x92, 0x5F, 0xFF, 0x06, 0x5C, 0xE7, 0xF7, 0x00, 0x00, 0x00, 0x00}, 9223372036854775807LL),    // 106'751'991 d + 14'454'775'807 us (maximum)
+    make_param({0xFF, 0xFF, 0xFF, 0xEF, 0x3F, 0xBA, 0xFF, 0xFF, 0x06, 0x5C, 0xE7, 0xF8, 0x00, 0x00, 0x00, 0x00}, 9223372036854775807LL),    // 106'751'992 d + -71'945'224'193 us (maximum)
+    make_param({0x7F, 0xFF, 0xFE, 0xD2, 0x40, 0x61, 0x5F, 0xFF, 0x21, 0x1D, 0x1A, 0xE3, 0xFE, 0xE5, 0x6E, 0x0A}, 9223372036854775807LL)     // -18'518'18 m + 555'555'555 d + 9'223'370'740'854'775'807 us (maximum)
+));
+
+struct recv_microseconds_overflow : recv, WithParamInterface<std::vector<char>> {
+    static_assert(std::chrono::microseconds::min().count() == std::numeric_limits<int64_t>::min());
+    static_assert(std::chrono::microseconds::max().count() == std::numeric_limits<int64_t>::max());
```
thed636: What are these assertions for?

include/ozo/ext/std/duration.h
thed636: I suggest using short alias for microseconds type like `us` to make shorter such sentences as `us::mix()`, `us::max()`.

tests/integration/result_integration.cpp
thed636: Sending the interval integration test is needed too.

tests/binary_deserialization.cpp
```diff
@@ -776,4 +776,78 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+};
+
+TEST_P(recv_microseconds_overflow, should_throw_if_value_too_big_or_to_small_to_be_stored_in_chrono_microseconds) {
```
thed636: Naming suggestion: 
`should_throw_if_value_too_big_or_to_small_to_be_stored_in_chrono_microseconds` -> `should_throw_if_value_out_of_range_of_chrono_microseconds`

tests/binary_deserialization.cpp
```diff
@@ -776,4 +776,78 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+    ozo::recv(value, oid_map, result);
+
+    EXPECT_EQ(result, expected);
+}
+
+inline auto make_param(std::array<unsigned char, 16> data, int64_t microseconds) {
+    std::vector<char> buffer(16, 0);
+    std::transform(data.cbegin(), data.cend(), buffer.begin(), [](auto c) { return static_cast<char>(c); });
+
+    return std::make_pair(buffer, std::chrono::microseconds(microseconds));
```
thed636: You have a copy of the `buffer` here. Boost.Ranges library is good for such a thing.

```c++
const auto to_signed_char = boost::adaptors::transformed([](unsigned char c) {
    return static_cast<char>(c); 
});

//...

auto buffer = data | to_signed_char;
return std::make_pair(std::vector<char>{buffer.begin(), buffer.end()}, 
    std::chrono::microseconds(microseconds));
```

thed636 (changes requested):

tests/binary_deserialization.cpp
```diff
@@ -776,4 +778,72 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+    std::chrono::microseconds result;
+    EXPECT_THROW(
+        ozo::recv(value, oid_map, result),
+        std::overflow_error
+    );
+}
+
+inline auto make_param(std::array<unsigned char, 16> data) {
+    std::vector<char> buffer(data.size(), 0);
+    std::transform(data.cbegin(), data.cend(), buffer.begin(), [](unsigned char c) { return static_cast<char>(c); });
```
thed636: Why to do not reuse transformed adaptor here?

tests/binary_deserialization.cpp
```diff
@@ -776,4 +778,72 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+
+    make_param({0xFF, 0xFF, 0xFF, 0xFC, 0xA2, 0x6D, 0xA0, 0x00, 0xF9, 0xA3, 0x18, 0x09, 0x00, 0x00, 0x00, 0x00}, -9223372036854775808ULL),  // -106'751'991 d + -14'454'775'808 us (minimum)
+    make_param({0x00, 0x00, 0x00, 0x10, 0xC0, 0x45, 0x00, 0x00, 0xF9, 0xA3, 0x18, 0x08, 0x00, 0x00, 0x00, 0x00}, -9223372036854775808ULL),  // -106'751'992 d + 71'945'224'192 us (minimum)
+    make_param({0xFF, 0xFF, 0xFF, 0x83, 0xEF, 0x61, 0x60, 0x00, 0xDE, 0xE2, 0xE5, 0x1D, 0x00, 0xE4, 0x45, 0xF7}, -9223372036854775808ULL),  // 14'960'119 m + -555'555'555 d + -532'854'775'808 us (minimum)
+
+    make_param({0x00, 0x00, 0x00, 0x03, 0x5D, 0x92, 0x5F, 0xFF, 0x06, 0x5C, 0xE7, 0xF7, 0x00, 0x00, 0x00, 0x00}, 9223372036854775807LL),    // 106'751'991 d + 14'454'775'807 us (maximum)
+    make_param({0xFF, 0xFF, 0xFF, 0xEF, 0x3F, 0xBA, 0xFF, 0xFF, 0x06, 0x5C, 0xE7, 0xF8, 0x00, 0x00, 0x00, 0x00}, 9223372036854775807LL),    // 106'751'992 d + -71'945'224'193 us (maximum)
+    make_param({0x7F, 0xFF, 0xFE, 0xD2, 0x40, 0x61, 0x5F, 0xFF, 0x21, 0x1D, 0x1A, 0xE3, 0xFE, 0xE5, 0x6E, 0x0A}, 9223372036854775807LL)     // -18'518'18 m + 555'555'555 d + 9'223'370'740'854'775'807 us (maximum)
```
thed636: You should choose between an assumption of microseconds representation and min/max values depended on the representation. These tests are not stable now.

BTW it is good practice to add values that help to determine which part of binary representation converted wrong.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,83 @@
...
+    }
+};
+
+template <>
+struct recv_impl<std::chrono::microseconds> {
+    template <typename OidMap>
+    static istream& apply(istream& in, size_type, const OidMap&, std::chrono::microseconds& out) {
+        static_assert(ozo::OidMap<OidMap>, "OidMap should model ozo::OidMap");
+
+        using microseconds = std::chrono::microseconds;
```
thed636: BTW `using std::chrono` inside the function gives you more readable code here.

tests/integration/request_integration.cpp
```diff
@@ -403,6 +403,26 @@ TEST(request, should_send_and_receive_empty_optional) {
...
 }
 
+TEST(request, should_send_and_receive_interval) {
+    using namespace ozo::literals;
+
+    const auto interval = ozo::pg::interval(239278272013014LL);
+
+    ozo::io_context io;
+    const ozo::connection_info<> conn_info(OZO_PG_TEST_CONNINFO);
+    const auto timeout = ozo::time_traits::duration::max();
```
thed636: No needs to make a timeout equal to `ozo::time_traits::duration::max()`, you can use the timeout-free version of the function.

thed636 (changes requested):

include/ozo/ext/std/duration.h
bmi-x: Not sure. It brings `[std::chrono::]days` and `[std::chrono::]months` into local namespace, but it already contains "my" `days` and `months`.
I think it's enough to add `using std::chrono::duration_cast` to reach almost the same readability.

tests/binary_deserialization.cpp
bmi-x: Tests for convertion by parts was added.
About min/max: what are your suggestions?

include/ozo/ext/std/duration.h
bmi-x: > The best solution is static assert on the representation type traits.

Now it seems to be fine for `std::chrono::microseconds::rep` up to about 103 bits length (value must be representable by int64_t days).

tests/binary_deserialization.cpp
thed636: IMO it is better to add static assertion for `std::chrono::microseconds::rep` type in `include/ozo/ext/std/duration.h`. A least it would not compile on platforms that are not supported by the tests.

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,96 @@
...
+struct send_impl<std::chrono::microseconds> {
+    template <typename OidMap>
+    static ostream& apply(ostream& out, const OidMap&, const std::chrono::microseconds& in) {
+        static_assert(ozo::OidMap<OidMap>, "OidMap should model ozo::OidMap");
+
+        static_assert(
+            std::chrono::microseconds::min().count() == std::numeric_limits<std::int64_t>::min() &&
+            std::chrono::microseconds::max().count() == std::numeric_limits<std::int64_t>::max(),
+            "std::chrono::microseconds tick representation type does not supported"
+        );
```
thed636: I think *"... is not supported"* is right.

tests/binary_deserialization.cpp
thed636: Oops, it is a bad idea. The best way is to separate reading data from the conversion algorithm. If you make something like 

```c++
namespace detail {
std::chrono::microseconds to_chrono_duration(const pg_interval& v) {
//...
}

} // namespace detail

template <>
struct recv_impl<std::chrono::microseconds> {
    template <typename OidMap>
    static istream& apply(istream& in, size_type, const OidMap&, std::chrono::microseconds& out) {
        //...
        detail::pg_interval interval;
        read(in, interval);
        out = detail::to_chrono_duration(interval);
    }
};
```

So you can test any corner cases you want without manual encoding them into a binary representation.

thed636 (changes requested):

tests/binary_deserialization.cpp
bmi-x: from_chrono_duration/to_chrono_duration algorithms are separated from reading and writing interval binary data.

tests/binary_deserialization.cpp
```diff
@@ -776,4 +778,68 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+
+    std::make_tuple(ozo::detail::pg_interval{  3333333333333333LL,          0,          0}, std::chrono::microseconds(3333333333333333LL)),
+    std::make_tuple(ozo::detail::pg_interval{                 0LL,     200000,          0}, std::chrono::microseconds(17280000000000000LL)),
+    std::make_tuple(ozo::detail::pg_interval{                 0LL,          0,       5555}, std::chrono::microseconds(14398560000000000LL)),
+
+    std::make_tuple(ozo::detail::pg_interval{      -14454775808LL,  -106751991,         0}, std::chrono::microseconds(-9223372036854775808ULL)), // minimum
+    std::make_tuple(ozo::detail::pg_interval{        71945224192LL, -106751992,         0}, std::chrono::microseconds(-9223372036854775808ULL)),
+    std::make_tuple(ozo::detail::pg_interval{      -532854775808LL, -555555555,  14960119}, std::chrono::microseconds(-9223372036854775808ULL)),
+
+    std::make_tuple(ozo::detail::pg_interval{        14454775807LL,  106751991,         0}, std::chrono::microseconds(9223372036854775807LL)), // maximum
```
thed636: `std::chrono::microseconds(9223372036854775807LL)` -> `std::chrono::microseconds::max()` ?

tests/binary_deserialization.cpp
```diff
@@ -776,4 +778,68 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+INSTANTIATE_TEST_CASE_P(convert_success, to_duration_ok, Values(
+    std::make_tuple(ozo::detail::pg_interval{       36672013014LL,         9,          92}, std::chrono::microseconds(239278272013014LL)),
+    std::make_tuple(ozo::detail::pg_interval{      -49727986986LL,       -20,          93}, std::chrono::microseconds(239278272013014LL)),
+    std::make_tuple(ozo::detail::pg_interval{   239278272013014LL,         0,           0}, std::chrono::microseconds(239278272013014LL)),
+
+    std::make_tuple(ozo::detail::pg_interval{  3333333333333333LL,          0,          0}, std::chrono::microseconds(3333333333333333LL)),
+    std::make_tuple(ozo::detail::pg_interval{                 0LL,     200000,          0}, std::chrono::microseconds(17280000000000000LL)),
+    std::make_tuple(ozo::detail::pg_interval{                 0LL,          0,       5555}, std::chrono::microseconds(14398560000000000LL)),
+
+    std::make_tuple(ozo::detail::pg_interval{      -14454775808LL,  -106751991,         0}, std::chrono::microseconds(-9223372036854775808ULL)), // minimum
```
thed636: `std::chrono::microseconds(-9223372036854775808ULL)` -> `std::chrono::microseconds::min()` ?
BTW why `ULL` ?

tests/binary_deserialization.cpp
```diff
@@ -776,4 +778,68 @@ TEST_F(recv, should_convert_TIMESTAMPOID_to_time_point) {
...
+
+TEST_P(to_duration_ok, should_convert_pg_interval_to_chrono_microseconds) {
+    const auto [interval, expected] = GetParam();
+
+    const auto result = ozo::detail::to_chrono_duration(interval);
+    EXPECT_EQ(result, expected);
+}
+
+INSTANTIATE_TEST_CASE_P(convert_success, to_duration_ok, Values(
+    std::make_tuple(ozo::detail::pg_interval{       36672013014LL,         9,          92}, std::chrono::microseconds(239278272013014LL)),
```
thed636: `std::chrono_literals` is very useful for constants: `std::chrono::microseconds(239278272013014LL)` -> [239278272013014us](https://en.cppreference.com/w/cpp/chrono/operator%22%22us).

include/ozo/ext/std/duration.h
```diff
@@ -0,0 +1,106 @@
...
+    using months = std::chrono::duration<std::int32_t, std::ratio<30 * days::period::num>>;
+
+    auto usecs_surplus = usecs(interval.microseconds) % days(1);
+    auto days_total = months(interval.months) + days(interval.days) + duration_cast<days>(usecs(interval.microseconds));
+
+    if (days_total < (duration_cast<days>(microseconds::min()) - days(1))
+        || ((days_total < days(0)) && ((days_total + days(1)) + usecs_surplus < microseconds::min() + days(1)))
+        || (duration_cast<days>(microseconds::max()) + days(1)) < days_total
+        || ((days(0) < days_total) && (microseconds::max() - days(1) < (days_total - days(1)) + usecs_surplus))) {
+        throw std::overflow_error("duration overflow: can not store interval value in chrono::duration");
```
thed636: I think it was a bad idea to throw an exception on overflow/underflow. I've just imaged how an unlucky client application started to crash because of somebody insert a row with an out of range value. Data bug with no chances to find it. So maybe it is better to set the variable into an appropriate limit value and to describe such behavior in the documentation?

thed636 (changes requested):

tests/binary_deserialization.cpp
bmi-x: `-X` is equal to `-(X)`, but `9223372036854775808` is out of range of `signed long long` (LLONG_MAX + 1).

thed636 (dismissed):

thed636 (approved):

PR #174, opened 2019-10-02
Fix usage for external projects

JonasProgrammer:
- find_package(ozo) will now resolve dependencies on it's own
- fixed include path when ozo is externally resolved
- moved installed tareget to yandex:: namespace
- vendored FindPostgreSQL of CMake 3.14
- added regression test to ensure ozo can be used from external projects

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ matrix:
 
   - os: linux
     dist: xenial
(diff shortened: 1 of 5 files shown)


thed636 (approved):

thed636:
Thanks a lot!

PR #175, opened 2019-10-03
Connection refine

thed636:
The Connection concept is fragile because of no consistency invariants protected by object methods.
From our experience, the customization points of Connection are not convenient and rarely usable. So it seems like the boost::asio::posix_descriptor interface is more feasible and convenient.
The new Connection interface looks like this:

```c++
struct connection {
    using native_handle_type = <implementation defined>;
    using oid_map_type = <implementation defined>;
    using error_context = <implementation defined>;
    using executor_type = <implementation defined>;

    native_handle_type native_handle() const noexcept;

    oid_map_type& oid_map() noexcept;
    const oid_map_type& oid_map() const noexcept;

    Statistics& statistics() noexcept;
    const Statistics& statistics() const noexcept;

    const error_context& get_error_context() const noexcept;
    void set_error_context(error_context v = error_context{});

    executor_type get_executor() const noexcept;
    error_code bind_executor(const executor_type& ex);

    error_code assign(native_conn_handle&& handle);

    template <typename WaitHandler>
    void async_wait_write(WaitHandler&& h) ;

    template <typename WaitHandler>
    void async_wait_read(WaitHandler&& h);

    error_code close() noexcept;

    void cancel() noexcept;
};
```


thed636:
No more needed

PR #176, opened 2019-10-08
Explicit connection implementation

thed636:
## Motivation

Connection in its current state is not a solid entity with a strong guarantee about consistency with a weak specification of data members. It is hard to support and it is hard to understand. The customization points had been shown as useless and wrong in terms of the absence of a solid concept guarantee.

## New concept

It is very similar to stream_descriptor from Asio, except the Connection allows changing its executor. In other cases, it led to clone-like interfaces that model more sophisticated behavior.

Here the description of the Connection concept from the documentation.

`Connection` concept represents a minimum set of attributes and functions that are required
by the library to establish communication and execute operations. `Connection` should provide:
* the native PostgreSQL connection handle from `libpq`,
* an executor to perform IO-related operation (according to the current version of Boost.Asio
  it should be `boost::asio::io_context::executor_type` object),
* an additional error context to provide context-depended information for errors,
* IO functions that are necessary to perform operations.

The default implementation of the concept is `ozo::connection`.

### Requirements

Any wrapper object, which may be unwrapped to the underlying Connection model via
`ozo::unwrap_connection` is valid Connection model.

Connection `c` is an object of type `C` for which these next requirements are valid:

| Expression | Type | Description |
|------------|------|-------------|
| <PRE>as_const(c).native_handle()</PRE> | `C::native_handle_type` | Should return native handle type of PostgreSQL connection. In the current implementation it should be `PGconn*` type. Shall not throw an exception. |
| <PRE>c.oid_map()<sup>[1]</sup><br/>as_const(c).oid_map()<sup>[2]</sup></PRE> | `C::oid_map_type` | Should return a reference<sup>[1]</sup> or a const reference<sup>[2]</sup> on `OidMap` which is used by the connection, the library may update this map during connection establishing. Shall not throw an exception. |
| <PRE>as_const(c).get_error_context()</PRE> | `C::error_context_type` | Should return a const reference on an additional error context is related to at least the last error. In the current implementation, the type supported is `std::string`. Shall not throw an exception. |
| <PRE>c.set_error_context(error_context)<sup>[1]</sup><br/>c.set_error_context()<sup>[2]</sup></PRE> | | Should set<sup>[1]</sup> or reset<sup>[2]</sup> additional error context. |
| <PRE>as_const(c).get_executor()</PRE> | `C::executor_type` | Should provide an executor object that is useful for IO-related operations, like timer and so on. In the current implementation `boost::asio::io_context::executor_type` is only applicable. Shall not throw an exception. |
| <PRE>c.set_executor(executor)</PRE> | `error_code` | Should change the executor for the specified one. This operation is used by `ozo::connection_pool` to provide connection migration between different instances of `boost::asio::io_service`. The call of the function during the active operation on connection is UB. The error should be indicated via the result. |
| <PRE>c.assign(move(native_conn_handle))</PRE> | `error_code` | Should assign `ozo::native_conn_handle` to the connection object. If success - the previous handle should be destroyed. The error should be indicated via the result. |
| <PRE>c.async_wait_write(WaitHandler)</PRE> | | Should asynchronously wait for write ready state of the connection socket. |
| <PRE>c.async_wait_read(WaitHandler)</PRE> | | Should asynchronously wait for read ready state of the connection socket. |
| <PRE>c.close()</PRE> | `error_code` | Should close connection socket and cancel all IO operation on the connection (like `async_wait_write`, `async_wait_read`). Shall not throw an exception. |
| <PRE>c.cancel()</PRE> | | Should cancel all IO operation on the connection (like `async_wait_write`, `async_wait_read`). Should not throw an exception. |
| <PRE>c.is_bad()</PRE> | bool | Should return `false` for the established connection that can perform operations. Shall not throw an exception. |
| <PRE>bool(as_const(c))</PRE> | bool | Should return `true` for the established connection that can perform operations. In fact it should be the negation of `c.is_bad()`. Shall not throw an exception. |
| <PRE>ozo::get_connection(c, t, Handler)</PRE> | | Should reset the additional error context. This behaviour is performed in default implementation of `ozo::get_connection()` via `c.set_error_context()` call. |
| <PRE>ozo::is_connection<C></PRE> | `std::true_type` | The template `ozo::is_connection` should be specialized for the connection type via inheritance from `std::true_type`. |

Where:
* `Handler` is callable with `template <typename Connection> void(ozo::error_code, Connection&&)` signature,
* `WaitHandler` is callable with `void(ozo::error_code, int = 0)` signature,
* `t` is a `TimeConstraint` model object,
* `as_const()` is `std::as_const()`,
* `move()` is `std::move()`.

### DIFF
--- a/include/ozo/detail/timeout_handler.h
+++ b/include/ozo/detail/timeout_handler.h
@@ -6,11 +6,11 @@
 namespace ozo::detail {
 
 template <typename Connection, typename Allocator>
-struct cancel_socket {
+struct cancel_io {
     Connection conn_;
     Allocator allocator_;
 
-    cancel_socket(Connection&& conn, const Allocator& allocator)
+    cancel_io(Connection&& conn, const Allocator& allocator)
     : conn_(std::forward<Connection>(conn)), allocator_(allocator) {}
 
     void operator() (error_code) const {
@@ -23,6 +23,6 @@ struct cancel_socket {
 };
 
 template <typename Connection, typename Allocator>
-cancel_socket(Connection&& conn, const Allocator& allocator) -> cancel_socket<Connection, Allocator>;
+cancel_io(Connection&& conn, const Allocator& allocator) -> cancel_io<Connection, Allocator>;
 
 } // namespace ozo::detail
--- a/include/ozo/detail/timeout_handler.h
+++ b/include/ozo/detail/timeout_handler.h
@@ -5,21 +5,24 @@
 
 namespace ozo::detail {
 
-template <typename Socket, typename Allocator>
+template <typename Connection, typename Allocator>
 struct cancel_socket {
-    Socket& socket_;
+    Connection conn_;
     Allocator allocator_;
 
-    cancel_socket(Socket& socket, const Allocator& allocator)
-    : socket_(socket), allocator_(allocator) {}
+    cancel_socket(Connection&& conn, const Allocator& allocator)
+    : conn_(std::forward<Connection>(conn)), allocator_(allocator) {}
 
-    void operator() (error_code ec) const {
-        socket_.cancel(ec);
+    void operator() (error_code) const {
+        conn_.cancel();
     }
 
     using allocator_type = Allocator;
 
     allocator_type get_allocator() const noexcept { return allocator_;}
 };
+
+template <typename Connection, typename Allocator>
+cancel_socket(Connection&& conn, const Allocator& allocator) -> cancel_socket<Connection, Allocator>;
 
 } // namespace ozo::detail
--- a/include/ozo/impl/connection.h
+++ b/include/ozo/impl/connection.h
@@ -159,10 +159,7 @@ inline error_code bind_executor(Connection& conn, const Executor& ex) {
 template <typename Connection>
 inline error_code close_connection(Connection&& conn) {
     static_assert(ozo::Connection<Connection>, "conn should model Connection");
-    error_code ec;
-    get_socket(conn).close(ec);
-    get_handle(conn).reset();
-    return ec;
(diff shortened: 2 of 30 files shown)


include/ozo/detail/timeout_handler.h
```diff
@@ -5,21 +5,24 @@
...
+template <typename Connection, typename Allocator>
 struct cancel_socket {
-    Socket& socket_;
+    Connection conn_;
     Allocator allocator_;
 
-    cancel_socket(Socket& socket, const Allocator& allocator)
-    : socket_(socket), allocator_(allocator) {}
+    cancel_socket(Connection&& conn, const Allocator& allocator)
+    : conn_(conn), allocator_(allocator) {}
```
bmi-x: std::move(conn)?

include/ozo/impl/connection.h
```diff
@@ -161,6 +161,24 @@ inline auto get_native_handle(const Connection& conn) noexcept {
     return unwrap_connection(conn).native_handle();
 }
 
+template <typename Connection>
+inline const auto& get_error_context(const Connection& conn) {
+    static_assert(ozo::Connection<Connection>, "T must be a Connection");
```
bmi-x: "T must be a Connection" -> "conn should model Connection"?

include/ozo/detail/timeout_handler.h
thed636: **Done**

Thanks, but `std::forward` is really needed.

include/ozo/impl/connection.h
thed636: **Done**

bmi-x (approved):

PR #183, opened 2019-10-09
Suppress -Wnewline-eof warning from Clang

jonesmz:
Change done by this script from stackoverflow: https://unix.stackexchange.com/questions/31947/how-to-add-a-newline-to-the-end-of-a-file

`git ls-files -z | while IFS= read -rd '' f; do tail -c1 < "$f" | read -r _ || echo >> "$f"; done`


thed636 (approved):
Thanks again for the PR!

PR #185, opened 2019-10-09
Fixes #177. Use of reserved identifier in macro name

### LINKED ISSUE #177: macro name is a reserved identifier
```
/home/jonesmz/meshpp_frameworks/3rdparty/ozo/include/ozo/error.h:527: warning: macro name is a reserved identifier [-Wreserved-id-macro]
        #define __OZO_SQLSTATE_NAME(value) case value: return std::string(#value) + "(" + detail::ltob36(value) + ")";
                ^
```

jonesmz:


thed636 (approved):
Thanks a lot for the PR!

PR #188, opened 2019-10-10
Add noexcept specifier to functions that do not throw exceptions

elsid:
Probably it's not the full list.


thed636 (approved):
Hi, Roma!

Nice to hear from you! Thanks for the PR!

Issue #181, opened 2019-10-09
Feature Request: Examples that don't use Boost.Coroutine


jonesmz:
Boost.Coroutine is, clearly, an extremely powerful framework. However, I am in a situation where adopting Boost.Coroutine is not currently an option.

I would like to see examples that don't use coroutines, that accomplish similar things to what's found here: https://github.com/yandex/ozo/blob/master/examples/request.cpp

For my situation, I am attempting to use OZO to replace libpqxx. Both OZO and libpqxx use the PostgreSQL provided "libpq" to run the underlying database operations, so they should be equally reliable and safe to use. The major difference, of course, is that OZO is asynchronous, and I've identified libpqxx's synchronous behavior as a major bottleneck in my application.

 My program is currently structured like this:

```cpp
// Header
struct ControllerV2
{
    ControllerV2(boost::asio::io_context & ioc, std::shared_ptr<pqxx::connection> pDbConn);
    boost::asio::io_context & m_ioc;
    std::shared_ptr<pqxx::connection> m_pDbConn;
}

int main()
{
    //boost program options parses CLI params, including connection string.
    std::shared_ptr<pqxx::connection> pDbConn = std::make_shared<pqxx::connection>(vm["db-connection-string"].as<std::string>());

    boost::asio::io_context ioc;    

    ControllerV2(ioc, std::move(pDbConn));

    ioc.run();

    return 0;
}

// controller.cpp
ControllerV2::ControllerV2(boost::asio::io_context & ioc, std::shared_ptr<pqxx::connection> pDbConn)
 : m_ioc(ioc)
 , m_pDbConn(std::move(pDbConn))
{
    // Do a bunch of stuff, including opening network connections, and registering callbacks via boost::asio for incoming network requests.
    // pDbConn gets passed around to other objects, such as various drivers for network connections for various protocols.
}
```

As you can see, I have some differences compared to the existing ozo examples.

1. An object that has a variety of subobjects that all do database stuff, instead of a single main() function.
2. Passing the database connection object into a non-template-type constructor (e.g., challange with ozo examples using "auto" for the connection info types.)
3. Need for lifetime management of the database connection object. E.g. value semantics? Pointer semantics?

So here's my wishlist of examples:

1. Less use of automatically deduced types (e.g. auto), because I don't want to convert all of my header file / cpp file pairs for various non-template classes / structs into template classes / structs.
[...]

PR #190, opened 2019-10-10
Fixes #187

### LINKED ISSUE #187: Implicit conversion loses integer precision
```cpp
template <typename T>
inline Require<Integral<T> && sizeof(T) == 1, istream&> read(istream& in, T& out) {
    out = in.get();
    if (!in) {
        throw system_error(error::unexpected_eof);
    }
    return in;
}
```

> ozo/include/ozo/impl/read.h:26: warning: implicit conversion loses integer precision: 'std::basic_istream<char, std::char_traits<char> >::int_type' (aka 'int') to 'char' [-Wimplicit-int-conversion]
>     out = in.get();
>         ~ ~~~^~~~~

> ozo/include/ozo/impl/read.h:51: in instantiation of function template specialization 'ozo::impl::read<char>' requested here
>     read(in, tmp);
>     ^

My suggestion is to use `boost::numeric_cast`

jonesmz:


include/ozo/impl/read.h
```diff
@@ -23,10 +23,11 @@ inline Require<RawDataWritable<T>, istream&> read(istream& in, T& out) {
 
 template <typename T>
 inline Require<Integral<T> && sizeof(T) == 1, istream&> read(istream& in, T& out) {
-    out = in.get();
-    if (!in) {
+    istream::traits_type::int_type c = in.get();
+    if (!in && istream::traits_type::not_eof(c)) {
```
thed636: EOF checking is not needed, `if (!in)` is enough. In other cases the functionality is broken - see the test that fails. So it looks like the best solution is:

```c++
in.get(out);
if (!in) {
    throw system_error(error::unexpected_eof);
}
return in;
```

thed636 (changes requested):

include/ozo/impl/read.h
jonesmz: Now all of the tests pass.

Did you still want the check for EOF removed?

thed636: Oops, I didn't figure out that the problem is the negation absence. Well, for now, I'd ask you to remove `EOF` checking, since it may change the throwing exception behaviour, which in case of an additional `EOF` checking depends on [std::basic_ios::exceptions()](https://en.cppreference.com/w/cpp/io/basic_ios/exceptions) method. It looks like a test case should be added to enforce such behavior.

It is clear that such behavior may hide the initial problem with the stream. For that case, the solution which throws a proper exception is needed. And IMO this is a subject for other PR.

jonesmz: Ok. I've removed the check for `not_eof(c)`

include/ozo/impl/read.h
```diff
@@ -23,10 +23,12 @@ inline Require<RawDataWritable<T>, istream&> read(istream& in, T& out) {
 
 template <typename T>
 inline Require<Integral<T> && sizeof(T) == 1, istream&> read(istream& in, T& out) {
-    out = in.get();
+    istream::traits_type::int_type c = in.get();
+    // TODO: Check result of istream::traits_type::not_eof(c)
```
thed636: Thanks, but could you please remove this comment and open an appropriate issue instead. I do not believe in TODO comments unfortunately they have no chances to be performed in real life in opposite to ticket or issues, which are always on the board.

Thank you!

thed636 (changes requested):

include/ozo/impl/read.h
jonesmz: I'm happy to work with you on this, but I think we're getting a little too deep into the minutia here.

You could have created the github issue that you want me to create, and merged this change as is (with or without the TODO: comment)

Instead, we've added another round trip between your team, and mine.

Note that I consider #187 a minor problem. I'm much more concerned with #181 . I'm still unable to use OZO to execute a query in my code. There are so many parts of OZO that are not easy to understand as an end user.

thed636 (approved):
Thanks a lot for the contribution and supporting the process! Sometimes it looks too annoying but it brings advantages in the future.

PR #193, opened 2019-10-17
Explicit connection cleanup

thed636:
Cleaning up rare used or unused entities.
Replace `hana::fold` to fold expression and `hana::for_ech` where possible to reduce compile-time.


include/ozo/connection.h
```diff
@@ -85,6 +85,7 @@ inline error_code bind_connection_executor(Connection&, const Executor&);
  * @brief Default model for `Connection` concept
  *
  * `Connection` concept model which is used by the library as default model.
+ * The class object is non-copyable.
```
serenityFX, line 88: This is weak relation with "Remove assign()" :) Please rename commit message

serenityFX (dismissed):

include/ozo/connection.h
thed636: **Done**

I've added some extra words to commit message about this.

serenityFX (approved):

PR #194, opened 2019-10-17
Fix deadlines handler

thed636:
## PROBLEM

Both callbacks (normal flow and cancel timer) appear in the io_context queue so both are called. No such proper handling for the situation.

## FIX

Add a shared state to prevent both callbacks execution. The backward atomic counter is used to indicate the calling sequence.
Error code for timed out operation now is `boost::asio::error::timed_out`.

## Add-on

Make `get_database`, `get_host`, `get_port`, `get_user`, `get_password` compatible with null handler

In `libpq` corresponding functions return `nullptr` for `nullptr` connection handle, `std::string_view` does not handle `nullptr` for `const char*` argument, so it leads to UB. This commit makes OZO functions to behave like the underlying `libpq` functions - return empty `std::string_view` for `nullptr` native connection handle.

### DIFF
--- a/include/ozo/detail/deadline.h
+++ b/include/ozo/detail/deadline.h
@@ -9,20 +9,25 @@
 
 namespace ozo::detail {
 
-template <typename Executor, typename Continuation>
+template <typename Executor, typename Continuation, typename TimerHandler>
 class deadline_handler {
 public:
-    template <typename TimeConstraint, typename TimerHandler>
+    template <typename TimeConstraint>
     deadline_handler(const Executor& ex, TimeConstraint t, Continuation handler, TimerHandler on_deadline)
-    : timer_(ozo::detail::get_operation_timer(ex, t)), handler_(std::move(handler)),
-            executor_(ozo::detail::make_strand_executor(ex)) {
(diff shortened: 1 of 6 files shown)


include/ozo/detail/deadline.h
```diff
@@ -33,32 +38,44 @@ class deadline_handler {
...
-        using allocator_type = asio::associated_allocator_t<Target>;
+        using allocator_type = deadline_handler::allocator_type;
 
-        allocator_type get_allocator() const noexcept { return asio::get_associated_allocator(target_);}
+        allocator_type get_allocator() const noexcept { return allocator_;}
 
         void operator() (error_code ec) {
-            if (ec != asio::error::operation_aborted) {
-                asio::dispatch(ozo::detail::bind(std::move(target_), std::move(ec)));
+            if (--(ctx_->state_) && ec != asio::error::operation_aborted) {
```
serenityFX: if ec = asio::error::operation_aborted call the wrong handler because state was decremented

thed636: Yep, thanks.
**Done**

serenityFX (approved):

PR #196, opened 2019-10-23
Refactoring of deadline timer handling and some other stuff

thed636:
## Refactor deadline timer handler

This C++Con'19 [talk](https://www.youtube.com/watch?v=3wy1OPqNZZ8) is about how it is right to create such handlers. The short story is - the last of operation handler or timer handler should call the client handler to provide a strong guarantee that no operation associated resources are used.

For the `ozo::cancel()` operation new solution is created. It is very similar to deleted `deadline_handler` and it can not guarantee that after the client handler call no operation associated resources are used because it can not be by the synchronous nature of underlying `libpq` `PQcancel()` implementation.

## Add-ons

* Reduce memory allocations for OIDs request while connecting. Because here we can use `std:array`. This is not a big deal but slightly reduces memory allocations during the retrieve of type OIDs. 
* Remove `async_connect_op` context. This entity is not needed since the `async_connect_op` is movable. Thus this reduces the number of memory allocations.
* Move result from shared context to the `async_get_result_op`. The result relates to the `async_get_result_op` only, so it is more correct to bind its life-time to this class object.
* Add result for PGRES_COMMAND_OK. PQresult contains additional information about command result, e.g., how many tuples were updated by the command, so previously a user just misses this information. Now it is possible to get it providing `ozo::basic_result` as a result destination for the request operation.
* Change `basic_result::handle()` to `basic_result::native_handle()` to be more conform with other classes containing underlying handles. Change documentation for the result to be more readable.
* Remove call_once as it is not needed anymore.
* Remove error-prone deadline_handler as it is not needed anymore.


include/ozo/impl/request_oid_map.h
```diff
@@ -51,41 +52,49 @@ inline void set_oid_map(oid_map_t<Impl>& oid_map, const oids_result& res) {
...
-    std::shared_ptr<oids_result> res_;
-
-    request_oid_map_op(Handler handler)
-    : handler_(std::move(handler)),
-      res_(std::allocate_shared<oids_result>(get_allocator())) {}
+    struct context {
+        Handler handler_;
+        oids_result res_;
+        context(Handler&& handler)
+        : handler_(std::move(handler)) {}
```
serenityFX, line 59: std::forward + decay?

include/ozo/detail/deadline.h
```diff
@@ -80,4 +80,82 @@ class deadline_handler {
...
+
+private:
+    using timer_type = typename ozo::detail::operation_timer<typename Stream::executor_type>::type;
+
+    struct context {
+        Stream& stream;
+        Handler handler;
+        Result result;
+        error_code ec;
+        int first_call{2};
```
serenityFX: why not first_call is bool? int value with first_call name looks weird and bad understanding.

include/ozo/impl/request_oid_map.h
thed636: Nope. `Handler&&` is an rvalue reference here, not the universal reference.

include/ozo/detail/deadline.h
thed636: Oops, there should be an atomic.
With an atomic variable, this is the only right way to work. It is a "standard" approach to manage invocation order in Boost.Asio and the same method is displayed in the presentation that is referred to in the PR description. So I decided to do not reinvent the wheel.
**DONE**

serenityFX (approved):

include/ozo/result.h
```diff
@@ -41,7 +42,8 @@ class value {
     }
 
     /**
-     * @brief Indicates text format representation for the value
+     * Determine whether the value is in binary format. Should be always false for current
```
bmi-x, line 45: may be "in _text_ format"?

PR #197, opened 2019-10-23
Reduce connection wrappers and unwrap stack

thed636:
### Problem

Connection in the pool is stored with a reference to last used io_context and active IO stream that attached to. The connection pool is io_context agnostic and is not served by any particular io_context. So it can be a dangling reference on io_context inside stored connection if the io_context object has been destroyed.

### The solution

The only right solution is to exclude `io_context` related entities from stored in the pool objects. Since there is a difference in the behavior of `connection` and `pooled_connection` it should be two different types.

The `pooled_connection` should not provide write access to oid_map, or allow to assign different connection handle. That's why some functionality has been removed from the `Connection` concept requirements.

Member function `connection::set_executor()` has been deleted, since no more such functionality is needed.

Add `is_open()` requirement as it is needed to determine the situation if the connection does not own the native handle.

Add `connection::release()` as it is needed to transfer the connection handle to different implementations. E.g., it is needed to implement a true io-context-unbound connection pool.

Using of `dup()` to make a duplicate of the connection file descriptor is removed in favor of `posix_descriptor::release()`. It should reduce the cost of the stream object attaching.

### DIFF
--- a/include/ozo/impl/connection.h
+++ b/include/ozo/impl/connection.h
@@ -12,22 +12,6 @@ namespace ozo {
 
 namespace detail {
 
-template <typename Connection, typename Executor>
-inline error_code bind_connection_executor(Connection& conn, const Executor& ex) {
-    if (conn.get_executor() != ex) {
-        typename Connection::stream_type s{ex.context()};
-        error_code ec;
-        s.assign(conn.socket_.native_handle(), ec);
-        if (ec) {
-            return ec;
-        }
-        conn.socket_.release();
-        conn.socket_ = std::move(s);
-        conn.io_ = std::addressof(ex.context());
-    }
-    return {};
-}
-
 template <typename NativeHandle>
 inline bool connection_status_bad(NativeHandle handle) noexcept {
     return !handle || PQstatus(handle) == CONNECTION_BAD;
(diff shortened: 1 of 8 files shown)


include/ozo/impl/connection_pool.h
```diff
@@ -11,44 +11,191 @@ namespace ozo::impl {
...
+
+    /**
+     * Get native connection handle object.
+     *
+     * This function may be used to obtain the underlying representation of the connection.
+     * This is intended to allow access to native `libpq` functionality that is not otherwise provided.
+     *
+     * @return native_handle_type --- native connection handle.
+     */
+    native_handle_type native_handle() const noexcept { return target().native_handle(); }
```
jonesmz: Not a reference?

thed636: Yep, it has `PGconn*` type in real life.

include/ozo/impl/connection.h
```diff
@@ -95,10 +94,8 @@ void connection<OidMap, Statistics>::async_wait_read(WaitHandler&& h) {
 
 template <typename OidMap, typename Statistics>
 error_code connection<OidMap, Statistics>::close() noexcept {
-    error_code ec;
-    socket_.close(ec);
```
yagzalex, line 99: Don't you need to still close the `socket_` thow?

thed636: Nope, since the socket is managed within `PGconn` object and will be closed right at the object destruction. Previously the socket_ object had a copy of connection socket file descriptor that obtained via `dup()` function so the `close()` call was necessary. Now there is no `dup()` call, so the `socket_` object should not manage the descriptor. But `socket_.release()` (that called inside the `connection::release()` function of the connection) will cancel all the IO on the `socket_`.

yagzalex (approved):

PR #200, opened 2019-10-29
Update tests infrastructure

thed636:
Test infrastructure is not the best thing here. So to continue with connection refactoring these changes are needed:

* **Update native connection handler mock to normal mock.** Now native connection handler is represented by the mock class, so now it is possible to test `libpq` calls without intermediate overloaded functions. Since `libpq` is a C library it is really easy to mock calls with SFINAE.
* **Add a stream descriptor dispatching from an Executor type.**  Like with the operation timer this approach allows customizing the IO stream type for testing purposes.
* **Make IO related mocks hosted by the test `io_context` class.**  Make `executor`, `strand_service`, `timer_service`, and `stream_service` mocks are hosted by the test `io_context` class to reduce boilerplate of test fixtures.

### DIFF
--- a/include/ozo/asio.h
+++ b/include/ozo/asio.h
@@ -4,6 +4,7 @@
 #include <boost/asio/executor.hpp>
 #include <boost/asio/strand.hpp>
 #include <boost/asio/steady_timer.hpp>
+#include <boost/asio/posix/stream_descriptor.hpp>
 
 namespace ozo {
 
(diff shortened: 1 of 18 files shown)


include/ozo/asio.h
```diff
@@ -4,6 +4,7 @@
 #include <boost/asio/executor.hpp>
```
serenityFX, line 4: > Like with the operation timer this approach allows to customize
the IO stream type for testing puposes

puRposes

serenityFX (approved):

Issue #201, opened 2019-10-30
ozo can't find resource_pool


mdyring:
(using cmake version 3.15.4)

Trying to use ozo on master branch and simply adding an `add_directory("vendor/ozo")` to my projects CMakeLists.txt.

This results in the below error:

```
[cmake] CMake Warning at vendor/ozo/CMakeLists.txt:14 (find_package):
[cmake]   By not providing "Findresource_pool.cmake" in CMAKE_MODULE_PATH this
[cmake]   project has asked CMake to find a package configuration file provided by
[cmake]   "resource_pool", but CMake did not find one.
[cmake] 
[cmake]   Could not find a package configuration file provided by "resource_pool"
[cmake]   (requested version 0.1.0) with any of the following names:
[cmake] 
[cmake]     resource_poolConfig.cmake
[cmake]     resource_pool-config.cmake
[cmake] 
[cmake]   Add the installation prefix of "resource_pool" to CMAKE_PREFIX_PATH or set
[cmake]   "resource_pool_DIR" to a directory containing one of the above files.  If
[cmake]   "resource_pool" provides a separate development package or SDK, be sure it
[cmake]   has been installed.
```

Only solution I've found for this is adding a QUIET to the find_package, alternatively commenting out the line to make it pick up the vendored resource_pool:

```
find_package(resource_pool 0.1.0 QUIET)
if (NOT resource_pool_FOUND)
    add_subdirectory(contrib)
endif()
```

JonasProgrammer:
Not an official ozo maintainer, but I am to blame for the very lines causing this.

First of all, this is not an error, just a warning, but I agree that it is ugly in the build logs.
I think your proposed solution of using `QUIET` is the right way to go. I saw you already created a PR, but you appear to have PRed to your own repository only; would you mind PRing it to the yandex repo as well?

And again, sorry for the inconvenience this might have caused you. I did not check for changes in the build output because all tests ran fine and the standalone/installed test worked fine as well.

thed636:
Hi, Martin!

Thanks for the PR.

Thanks for the help, Jonas!

thed636:
The PR is successfully merged, so I'm closing the issue.

@JonasProgrammer  @mdyring 
Jonas, Martin, thank you one more time!

PR #202, opened 2019-10-30
Don't fail immediately if resource_pool is missing

mdyring:
In reference to https://github.com/yandex/ozo/issues/201


thed636 (approved):
PR will be merged right after #197

PR #207, opened 2019-11-10
Make begin_op::with_transaction_options constexpr

JonasProgrammer:
When implementing transaction options, I forgot to make `with_transaction_options` `constexpr`. Given that both `begin_op`'s constructor and `get_operation_initiator` are in fact `constexpr`, there is no reason it should not be; sorry for that oversight.

As a side question, is there a specific place for 'compile time tests' or something? It would make sense to have something stop compile, should the interface ever change to lose its `constexpr`ness for some reason, but there is no test for `begin_op` specifically (and creating a new file for just one test felt strange); the transaction integration test on the other hand is only an 'opt-in test' and therefore not really the place to test compile time stuff.

### DIFF
--- a/include/ozo/transaction.h
+++ b/include/ozo/transaction.h
@@ -35,7 +35,7 @@ struct begin_op : base_async_operation <begin_op<Initiator, Options>, Initiator>
     }
 
     template <typename OtherOptions>
-    auto with_transaction_options(const OtherOptions& options) const {
+    constexpr auto with_transaction_options(const OtherOptions& options) const {
         return begin_op<Initiator, OtherOptions>{get_operation_initiator(*this), options};
     }
 


thed636 (approved):

thed636:
> As a side question, is there a specific place for 'compile time tests' or something? It would make sense to have something stop compile, should the interface ever change to lose its constexprness for some reason, but there is no test for begin_op specifically (and creating a new file for just one test felt strange); the transaction integration test on the other hand is only an 'opt-in test' and therefore not really the place to test compile time stuff.

Well, it is ok to create `tests/transaction.cpp` and place your desired test into this file. The `transaction` refactoring is on the way, so this file will be filled with additional tests. I didn't think about compile-time tests so much. We have several static assertions about concept models, but that is all. From one side, it looks like to test constants or something immutable, but if you see an outcome from such an approach, try it.

thed636:
PS: thanks for the PR, Jonas!

JonasProgrammer:
Ok, I'll just create said file then and put a `constexpr` definition in the anonymous namespace, so it does not clash with anything. To my knowledge, there is no way to test for `constexpr`ness in terms of a type trait in C++17, so I don't think there is any good way to write real test code for that.
It would just be nice to have this covered somewhere, so nobody accidentally breaks client code that relies on this behaviour by just forgetting the keyword (like I did) -- unless of course it is really necessary; yandex is the first-class customer, after all.

PR #212, opened 2019-11-15
Use implicit deduction guides

elsid:
Avoid using unnecessary functions like `make_*` that are only required for type deduction. C++17 provides the same feature for constructors.

### DIFF
--- a/include/ozo/type_traits.h
+++ b/include/ozo/type_traits.h
@@ -522,7 +522,7 @@ OZO_PG_DEFINE_CUSTOM_TYPE(custom_type, "code.custom_type")
 
 //...
 // Creating ConnectionSource for futher requests to a database
-const auto conn_source = ozo::make_connection_info("...", regiter_types<custom_type>());
+const ozo::connection_info conn_source("...", register_types<custom_type>());
  * @endcode
  *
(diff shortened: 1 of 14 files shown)


thed636 (approved):

include/ozo/type_traits.h
```diff
@@ -522,7 +522,7 @@ OZO_PG_DEFINE_CUSTOM_TYPE(custom_type, "code.custom_type")
 
 //...
 // Creating ConnectionSource for futher requests to a database
-const auto conn_source = ozo::make_connection_info("...", regiter_types<custom_type>());
```
jonesmz, line 525: regiter_types == register_types ?

thed636: Thanks, Michael!

Roma, could you fix this typo, please?

thed636 (changes requested):
GitHub does not allow us to change the resolution without comments. :)

include/ozo/type_traits.h
elsid: Done

thed636: Thanks a lot!

thed636 (approved):

PR #213, opened 2019-11-24
Support thread safety customization for connection_pool

elsid:
Allows to avoid useless synchronization on mutex in single threaded applications. Requires to update resource_pool version to be able uncouple handle and pool types. Default connection_pool still have synchronization. New type alias `single_threaded_connection_pool` removes synchronization.

https://github.com/elsid/resource_pool/pull/19 is required to be merged be before this pr to avoid dependency to a branch.


jonesmz:
Consider "non-locking" instead of single-threaded, as it should still be usable in multiple threads, if the consuming application structures itself properly.

thed636:
Hi, Roma!

I think the better solution is to operate not in terms of mutex type, but terms of a thread-safety trait. IMO OZO shall not expose the implementation details of the `resource_pool` library. So, in that case, we should have something like this:

```c++

template <typename V>
struct thread_safity : V {
    using base = V;

    constexpr auto operator !() const noexcept {
        return thread_safity<typename std::negation<V>::type>{};
    }
};

constexpr thread_safity <std::true_type> thread_safe;

template <typename Source, typename ThreadSafety = decltype(thread_safe)>
class connection_pool {
public:
    using connection_rep_type = ozo::connection_rep<typename ozo::unwrap_type<ozo::connection_type<Source>>::oid_map_type>;
    
    static constexpr ThreadSafety is_thread_safe;

    using impl_type = detail::get_connection_pool_impl_t<connection_rep_type, ThreadSafety>;

    connection_pool(Source source, const connection_pool_config& config, const ThreadSafety& = ThreadSafety{});
//...
};
```

So a user can explicitly point the desired thread-safety in the constructor:

```c++
auto thread_safe_pool = ozo:: connection_pool{connection_source, thread_safe};

auto relaxed_pool = ozo:: connection_pool{connection_source, !thread_safe};
```

@jonesmz Michael, what do you think about it?

include/ozo/thread_safety.h
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @brief defines admissibility to use in multithreaded environment without additional synchronization
+ *
+ * Represents binary state. True enables memory sychronization for underlying types to support
+ * safe access in multithreaded environment. False disables memory synchronization.
+ *
+ * @tparam enabled --- binary state.
+ */
+template <bool enabled>
+struct thread_safety : std::integral_constant<bool, enabled> {
```
thed636: May be better

```c++
std::bool_constant<enabled>
```
?

include/ozo/thread_safety.h
```diff
@@ -0,0 +1,26 @@
...
+    using base_type = std::integral_constant<bool, enabled>;
```
thed636: What is the purpose of `base_type`?

tests/integration/connection_pool_integration.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+namespace {
+
+namespace asio = boost::asio;
+
+using namespace testing;
+
+TEST(connection_pool, get_connection_twice_should_get_the_same) {
```
thed636: It is better to choose  `connection_pool_integration` name to avoid conflict with other `connection_pool` tests. Since we link all the tests in a single binary we, unfortunately, have such a problem.

thed636 (changes requested):

include/ozo/thread_safety.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
thed636: It is better to move the header into the `core` folder

include/ozo/thread_safety.h
```diff
@@ -0,0 +1,26 @@
...
+#include <type_traits>
+
+namespace ozo {
+
+/**
+ * @brief defines admissibility to use in multithreaded environment without additional synchronization
+ *
+ * Represents binary state. True enables memory sychronization for underlying types to support
+ * safe access in multithreaded environment. False disables memory synchronization.
+ *
```
thed636:
```
@ingroup group-core-types
```

include/ozo/connection_pool.h
```diff
@@ -294,21 +295,22 @@ struct connection_traits<yamail::resource_pool::handle<Pool>> :
...
     using connection_rep_type = ozo::connection_rep<typename ozo::unwrap_type<ozo::connection_type<Source>>::oid_map_type>;
 
-    using impl_type = yamail::resource_pool::async::pool<connection_rep_type>;
+    using impl_type = detail::get_connection_pool_impl_t<connection_rep_type, ThreadSafety>;
     /**
      * @brief Construct a new connection pool object
      *
      * @param source --- #ConnectionSource object which is being used to create connection to a database.
      * @param config --- pool configuration.
+     * @param thread_safety -- admissibility to use in multithreaded environment without additional synchronization.
```
thed636, line 313: Missing default value documentation

include/ozo/connection_pool.h
```diff
@@ -393,11 +395,13 @@ constexpr auto ConnectionPool = is_connection_pool<std::decay_t<T>>::value;
  *
  * @param source --- #ConnectionSource object which is being used to create connection to a database.
  * @param config --- pool configuration.
+ * @param thread_safety --- admissibility to use in multithreaded environment without additional synchronization.
```
thed636, line 394: Missing default value documentation

thed636 (changes requested):

include/ozo/thread_safety.h
elsid: Done

include/ozo/thread_safety.h
elsid: None, forgot to remove.

tests/integration/connection_pool_integration.cpp
elsid: Done

include/ozo/thread_safety.h
elsid: Done

include/ozo/thread_safety.h
elsid: Done

include/ozo/connection_pool.h
elsid: Done

include/ozo/connection_pool.h
elsid: Done

thed636 (approved):

Issue #215, opened 2019-11-27
CMake Error: `resource_pool` is missing.


ricejasonf:
```
CMake Error at contrib/CMakeLists.txt:1 (add_subdirectory):
  The source directory

    /usr/local/src/ozo-4bf9f42098cead27d4a3d4c1af2d06eb360f24cb/contrib/resource_pool

  does not contain a CMakeLists.txt file.
```

This happens when building from an archive of the source code (ie without git).

This is a bit of a speed bump for people trying use this library for the first time. Perhaps documenting `resource_pool` as a dependency would be useful.

Is it a required dependency?

thed636:
Hi, Jason!

Yes, `resource_pool` is required, and you're right, it should be documented. Thanks for the report!

Issue #203, opened 2019-11-01
Replace BOOST_FUSION_ADAPT_STRUCT with BOOST_HANA_ADAPT_STRUCT in examples


jonesmz:
Since ozo already explicitly depends on Boost.Hana, I think it makes sense to provide examples for adapting structs that use Boost.Hana instead of Boost.Fusion.

My motivation behind recommending this is actually that I ran into a problem including BOOST_FUSION_ADAPT_STRUCT in the same translation unit as a Boost.MultiIndex container. It causes a compiler error, no idea why. However, BOOST_HANA_ADAPT_STRUCT doesn't have the same problem.

thed636:
Hi Michael!

Could you provide additional information about of problem do you have? We are using both Boost.Fusion and Boost.MultiIndex in many translation units and have no problem.

jonesmz:
See this discussion for more information: https://github.com/redboltz/mqtt_cpp/issues/351

*Appears* to be the same problem that I had when I opened this issue, but the error messages were different, and given C++'s errors tend to be unworkable / unreadable  sometimes, I just didn't try to track down the underlying issue.

Using BOOST_HANA_ADAPT_STRUCT instead of BOOST_FUSION_ADAPT_STRUCT allowed me to compile OZO into my application.

thed636:
Hi, Michael! 

Thanks a lot for the information! I think, first of all, I'll add some references for this problem into the documentation. According to examples - I think at least one of them should use Boost.Fusion just for the display purpose.

thed636:
Documentation and examples are updated.

PR #216, opened 2019-12-04
Use Boost.Hana as a primary mechanism for structures adaptation 

thed636:
#203 - replacing `BOOST_FUSION_ADAPT_STRUCT` with `BOOST_HANA_ADAPT_STRUCT`.
#215 - add resource_pool to the dependency list in README.md

### DIFF
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -3,12 +3,12 @@
 Here are some examples of how to use OZO API.
 
 <!-- TOC -->
-- [How to](#How-to)
-  - [How To Make A Very Simple Request](#How-To-Make-A-Very-Simple-Request)
-  - [How To Handle Error Properly](#How-To-Handle-Error-Properly)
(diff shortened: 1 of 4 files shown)


docs/howto.md
```diff
@@ -198,6 +200,8 @@ int main() {
 
 In this example, you can change the order of the fields in either `my_row` or your database query freely, as the fields are identified based on their name, and not based on their index in the tuple.
 
+ > **Note:** you may use `BOOST_FUSION_ADAPT_STRUCT` and all of the `Boost.Fusion` adaptation mechanisms for structures, but it is recommended to use `Boost.Hana` so as a faster and modern approach.
```
jonesmz: > but it is recommended to use `Boost.Hana` so as a faster and modern approach.

but it is recommended to use `Boost.Hana` for a faster and more modern approach.

thed636: Thanks, Michael!

massaraksh111 (approved):

serenityFX (approved):

PR #218, opened 2019-12-24
Transaction connection

thed636:
Transaction as Connection model

`Connection` concept model that describes the open transaction on the underlying connection. The transaction is *not* an RAII object it does not control transaction flow. It is dedicated to describe the current transaction and represent a transaction on the type level.

Depending on the underlying `Connection` the transaction object may model `Nullable`. If the object is in null state call of any member may lead to undefined behaviour.


include/ozo/transaction.h
```diff
@@ -136,8 +583,10 @@ constexpr auto get_transaction_mode(const impl::transaction<Ts...>& transaction)
...
  */
 template<typename... Ts>
-constexpr auto get_transaction_deferrability(const impl::transaction<Ts...>& transaction) {
+constexpr auto get_transaction_deferrability(const transaction<Ts...>& transaction) {
     return get_option(transaction.options(), transaction_options::deferrability, none);
 }
 
 } // namespace ozo
+
+#include <ozo/impl/transaction.h>
```
elsid, line 592: Is it possible to move transaction type to a different header to avoid this cyclic include dependecy?

elsid (changes requested):

include/ozo/transaction.h
thed636: Well, actually there is no problem with circular dependency since the implementation header is included at the end of public header. And the public header is included at the begin of the implementation header. So in any case we have the same sequence of included headers and every header included only once.

elsid (approved):

PR #225, opened 2019-12-30
Fixes #223 : Change spelling of is_time_constrain to is_time_constraint

### LINKED ISSUE #223: Confusing type trait spelling : is_time_constrain
I feel like this should instead be named is_time_constraint.

```cpp
template <typename T>
struct is_time_constrain : std::false_type {};

/**
 * @brief Time constrain concept
 *
 * `TimeConstraint` is a type which provides information about time restrictions for an operation.
 * Currently supported constrains:
 * * `std::chrono::duration` --- operation time-out duration,
 * * `std::chrono::time_point` --- operation deadline time point,
 * * `ozo::none` --- operation is not restricted in time.
 */
template <typename T>
inline constexpr auto TimeConstraint = is_time_constrain<std::decay_t<T>>::value;
```

jonesmz:


thed636 (approved):
Thanks for the PR!

PR #229, opened 2020-01-09
Update resource_pool third party dependency

jonesmz:
resource_pool has several warning fixes that I would enjoy having in ozo, so that I can have quieter builds.

Please let me know if updating to the current master branch is inappropriate in some way, maybe we can work out an alternative way to handle this.


thed636 (approved):
Thanks for the PR!

PR #240, opened 2020-03-27
Add share ability for ozo::result

### LINKED ISSUE #236: Stubbing request
It is very desirable to have a possibility of stubbing or mocking database requests for tests and as the request is a template it is not an easy task.

Is it any recommended way to do this by yandex?

After some research, I found 2 ways to do this:
1) Write a specialized interface for every specialization of method request what I will need to use.

2)Write a wrapper that takes std::Function object as request implementation. 
```
template <typename Reqfun> class DbManager {
public:
    explicit DbManager(Reqfun fun) : reqImplFun(fun) {
    }

    DbManager(DbManager &) = delete;

    template <typename ConnectionProvider, typename BinaryQueryConvertible, typename TimeConstraint, typename Out,
              typename CompletionToken>
    decltype(auto) request(ConnectionProvider &&provider, BinaryQueryConvertible &&query, TimeConstraint time_constraint, Out out,
                           CompletionToken &&token) {
        return reqImplFun(provider, query, time_constraint, out, token);
    }

private:
    Reqfun reqImplFun;
};
```
Then it can be passed wrapped by lambda ozo::request for production code
```
[](auto &&provider, auto &&query, auto time_constraint, auto out, auto &&token) {
    return ozo::request(provider, query, time_constraint, out, token);
};
```
or stub implementation for tests.

Do you have any better solutions than these above?

thed636:
Initially `ozo::result` is not copyable object, but sometimes it is better to have copyable result object. This is solved via `ozo::shared_result` object which may be copyable. Shared result may be obtained via `ozo::result::share()` member function.

### DIFF
--- a/include/ozo/result.h
+++ b/include/ozo/result.h
@@ -316,14 +316,22 @@ public:
 
     basic_result() = default;
     basic_result(handle_type res) noexcept(noexcept(handle_type(std::move(res))))
-    : res_(std::move(res)) {}
+    : handle_(std::move(res)) {}
+
+    template <typename Other, typename = hana::when<
+        hana::is_convertible<Other, T>::value && !std::is_same_v<T, Other>
+    >>
+    basic_result(basic_result<Other>&& x) noexcept(
+        noexcept(handle_type{hana::to<T>(std::move(x.release()))}))
+    : handle_(hana::to<T>(std::move(x.release()))) {
+    }
 
     /**
      * Iterator on the first row of the result.
      *
      * @return const_iterator --- iterator on the first row
      */
-    const_iterator begin() const noexcept { return {{std::addressof(*res_), 0, 0}}; }
+    const_iterator begin() const noexcept { return {{native_handle(), 0, 0}}; }
 
     /**
      * Iterator on end of row sequence.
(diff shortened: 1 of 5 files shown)


include/ozo/result.h
```diff
@@ -391,13 +391,40 @@ class basic_result {
...
+ *
+ * @code
+ozo::result res;
+//...
+share_result sres = res.share();
+ * @endcode
+ *
+ * @ingroup group-requests-types
+ */
+using shared_result = basic_result<pg::shared_safe_handle_t<pg::result>>;
```
elsid: Should we add tests for a new API type?

include/ozo/result.h
```diff
@@ -406,7 +433,18 @@ class basic_result {
...
+public:
+    using basic_result<pg::result>::basic_result;
+    /**
+     * Transfers the underlying handle, if any, to a `ozo::shared_result` object. Multiple
+     * `ozo::shared_result` objects may reference the result handle, which is not possible
+     * with `ozo::result`. After calling `share()` on a `ozo::result`, `valid() == false`.
+     *
+     * @return shared_result_type - shared result object
+     */
+    shared_result share() { return pg::share_handle(std::move(handle()));}
```
elsid: Intuitively, a method named like this does not take ownership of underlying value. How about to use something like `release_shared`?

elsid (changes requested):

include/ozo/result.h
thed636: Well, I did it in a similar way as [std::future::share](https://en.cppreference.com/w/cpp/thread/future/share) just to conform to standard things with similar semantic.

include/ozo/result.h
thed636: Well, we already have [ozo::basic_result tests](https://github.com/yandex/ozo/blob/master/tests/result.cpp#L184) that cover all the functionality but the integration tests certainly should be added. Thanks for the notice!

elsid (approved):

PR #246, opened 2020-04-07
Support benchmark parameters as process args and different output formats

elsid:
* General feature is ability to run specific benchmark using command line with output into JSON that can be used by other program (e.g. make a plot).
* New dependencies to nlohmann/json and Boost.Program_options are added.
* Single threaded benchmarks are using connection pool without synchronization.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -8,6 +8,8 @@ endif()
 add_executable(ozo_benchmark ozo_benchmark.cpp)
 target_link_libraries(ozo_benchmark ozo)
 
+find_package(Boost COMPONENTS coroutine context system thread atomic program_options REQUIRED)
+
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -2,3 +2,7 @@
 	path = contrib/resource_pool
 	url = https://github.com/elsid/resource_pool.git
 	branch = master
+[submodule "contrib/nlohmann_json"]
```
thed636: Is it better to get `nlohmann_json` as an external project and check out it only for benchmarks target in the same way as `gtest`?

thed636 (changes requested):

.gitmodules
elsid: Done

thed636 (approved):

PR #248, opened 2020-04-07
Fix test instructions

jonesmz:


jonesmz:
Another change I could make a PR for is to modify

```
make -j$(nproc)
```
to
```
cmake --build . --parallel $(nproc) --verbose --clean-first
```

Which is a more idomatic way to build cmake projects.

thed636 (approved):

jonesmz:
Do you want me to add another commit to change it to `cmake --build . --parallel $(nproc) --verbose --clean-first` ?

Or would you rather leave it as `make` ?

thed636:
@jonesmz would it be nice to add it as a separate recommendation to simplify and speed up the process?

jonesmz:
The reason someone might want "cmake --build ." is because they won't need to know anything about the generator that cmake defaulted to on the current platform.

Theres:

- Make
- MSBuild
- Ninja

and others

thed636:
Yes, that’s a good reason, let’s add it instead of `make`.

jonesmz:
Just to confirm you're looking to have me make a PR that adds cmake --build . --parallel $(nproc) --verbose --clean-first to the instructions?

thed636:
Yes, sure. I just merged this PR, since it was approved.

PR #249, opened 2020-04-07
CMakeLists should not know anything about ccache

jonesmz:
Support for CMAKE_C_COMPILER_LAUNCHER was added in CMake 3.4

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -112,6 +112,8 @@ build() {
     mkdir -p ${BUILD_DIR}
     cd ${BUILD_DIR}
     cmake \
+        -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+        -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
         -DCMAKE_C_COMPILER="${CC_COMPILER}" \
         -DCMAKE_CXX_COMPILER="${CXX_COMPILER}" \
         -DCMAKE_CXX_FLAGS="$CMAKE_CXX_FLAGS"\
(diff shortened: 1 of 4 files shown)


scripts/build.sh
```diff
@@ -112,6 +112,8 @@ build() {
     mkdir -p ${BUILD_DIR}
     cd ${BUILD_DIR}
     cmake \
+        -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+        -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
```
thed636, line 116: It looks like to conform to current behaviour there should be check for the `ccache` existence. And it would be nice to allow a user to specify a launcher, e.g. maybe somebody wants to use `distcc` or something custom.

thed636 (changes requested):

thed636:
Hi Michael!

Thanks for the participation! It would be nice to hear from you why `CMakeLists should not know anything about ccache`.

jonesmz:
I have ccache installed on my machine, but don't want the ozo examples to use it while being built.

Automatically discovering the availability of a program, and modifying the compilation process based on that automatic discovery causes surprises for the end-user of the code. It further makes it more difficult for linux distribution packagers, and people trying to consume the code via cmake include subdirectory statements.

thed636:
That's a good point!

scripts/build.sh
jonesmz: Ok, this is a reasonable request.

I won't be able to get to this today, but sometime this week, probably.

PR #250, opened 2020-04-08
Add github actions as a second continuous integration platform

jonesmz:
The reason why this is beneficial to people is because github actions are included in all github accounts, where configuring a continuous integration service like travis CI is not super trivial to do. It's obviously not impossible, since I did it once upon a time, but it's kind of a pain in the rear nevertheless.

So people working on a fork of OZO on github, even if they don't have travis CI, will automatically have their work run through continuous integration, even before they create a pull request for the main OZO repository.

Further, github actions allow up to 20 parallel jobs (5 for macos), which can allow the OZO continuous integration to complete significantly faster.

And lastly, in my experience, on a job per job basis, it's been my experience that github actions are quite a bit faster than travis ci.

Note: This PR includes https://github.com/yandex/ozo/pull/249, and then removes ccache usage entirely. This is because the github CI is quite a bit faster than travis, while also having a different security model than travis, making the existing ccaching code more complicated to use. I'd be happy to leave that in if someone can point out how to adjust the github actions environment to be happy with it.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,37 @@
+name: Linux
+
+on:
+  pull_request:
+  push:
+  release:
+    types: published
+
+jobs:
+  linux:
+    runs-on: [ubuntu-latest]
+    strategy:
+      matrix:
+        pattern:
+          - 'pg docker gcc debug'
+          - 'pg docker gcc release'
+          - 'pg docker gcc test_external_project'
+          - 'pg docker clang debug'
+          - 'pg docker clang release'
+          - 'pg docker clang asan'
+          - 'pg docker clang ubsan'
+          - 'pg docker clang tsan'
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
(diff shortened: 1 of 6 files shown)


elsid:
ccache is used not only for CI. Current configuraiton allows to use it locally. I suggest to add `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER` to CMake args in `scripts/build.sh` to allow custom usage of ccache in the most simple way through environment variables:
```bash
cmake \
        -DCMAKE_C_COMPILER_LAUNCHER="${OZO_CMAKE_C_COMPILER_LAUNCHER}" \
        -DCMAKE_CXX_COMPILER_LAUNCHER="${OZO_CMAKE_CXX_COMPILER_LAUNCHER}" \
        ...
```
with expectation to be used like this:
```bash
env OZO_CMAKE_C_COMPILER_LAUNCHER=ccache OZO_CMAKE_CXX_COMPILER_LAUNCHER=ccache scripts/build.sh
```
and by default it will be disabled.

PR #256, opened 2020-05-02
Support connection pool invalidation

elsid:


include/ozo/connection_pool.h
```diff
@@ -344,6 +344,13 @@ class connection_pool {
         return connection_provider(*this, io);
     }
 
+    /**
+     * Invalidate all contained connections. It will prevent all available and currently used connections to be reused.
```
thed636, line 348: Could you please to add a use-case description for this function?

thed636 (changes requested):

include/ozo/connection_pool.h
elsid: I can think of something like: close all connections when it's considered that all of them are invalid due to some user defined heuristics but the only way to truly check it is to wait for each connection until request is timed out.

thed636: Well, it's maybe a good point. E.g. invalidate all connections if get some wrong OID error or something like this. I mean, the feature should have a particular use case or nobody will use it. Another use case - invalidate all the connections by a signal like a log rotation.

thed636 (changes requested):

PR #260, opened 2020-06-15
Add constexpr version of ozo::detail::ltob36

massaraksh111:


include/ozo/detail/base36.h
```diff
@@ -20,7 +44,8 @@ inline std::string ltob36(long i) {
         u /= base;
     } while (u > 0);
 
-    boost::range::reverse(out);
+    reverse(out);
```
jonesmz, line 47: what's wrong with boost::range::reverse?

PR #268, opened 2020-11-22
Custom binary query

### LINKED ISSUE #267: How to build INSERT query with multiple rows?
Postgres supports one INSERT statement has multiple values / rows.

How can I build such query in ozo given that I don't know the number of rows in advance?

INSERT INTO test (a, b) VALUES (1, 2) (3, 4) (5, 6);

elsid:
Make binary query implementation interface public to allow custom implementation. And provide a usage example. Primarily resolve https://github.com/yandex/ozo/issues/267.

### DIFF
--- a/include/ozo/io/binary_query.h
+++ b/include/ozo/io/binary_query.h
@@ -36,8 +36,57 @@ namespace ozo {
  */
 class binary_query {
 public:
+    static constexpr auto binary_format = 1;
+
+    struct implementation {
+        /**
+          * Get raw query text buffer.
+          *
+          * @return `const char*` --- pointer to the buffer.
+          */
+        virtual const char* text() const noexcept = 0;
+
+        /**
+          * Get query parameter types array.
+          *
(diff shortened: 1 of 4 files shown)


include/ozo/io/binary_query.h
```diff
@@ -36,8 +36,57 @@ namespace ozo {
  */
 class binary_query {
 public:
+    static constexpr auto binary_format = 1;
+
+    struct implementation {
```
thed636, line 41: I'd suggest to do not expose the interface, since the inheritance is a very tight coupling relation that is not necessary for this case and restricts us from any direct changes of the internals. 

The concept mechanism with type-erasure is much preferable in this case.

If the `binary_query` becomes a type-erasure type for concept models support -  it should not construct any implementation by-default, the implementation of the fixed-number parameters should be separated.

thed636 (changes requested):
The library is concept-based, so the polymorphism should be implemented via concepts first.

Issue #280, opened 2021-01-23
CMakeLists.txt generates unstable package target


JonasProgrammer:
Hi,

I stumbled upon a minor issue with the current `CMakeLists.txt` when reviewing generated package files: The `ozo` CMake targets being exported from the library conditionally define a couple of warning flags, depending on the compiler used for packaging.
[These lines](https://github.com/yandex/ozo/blob/8a6597623816cbda63f329ff765ba2db95059c88/CMakeLists.txt#L30) are responsible for this.

As to why this is an issue: First of all, including warning flags in exported targets at all is a debatable practise. Second, despite being header-only, the packaged version's cmake files (to be consumed by external projects) slightly differ, depending on which compiler was used *during packaging* -- with the warnings being included, if clang was present or absent, if any other compiler was used.

This is not an issue when you use ozo as a submodule or pull it in via `ExternalProject_add`, which is what you do at yandex, I presume. Still, for the use case of an external package, a stable build output regardless of the compiler present on the packaging system (if any) would be desirable, so it can be copied and consumed to other system configurations.

I think simply putting the offending options in a generator expression, so it is only picked up during build time, and then mirroring the behaviour in `cmake/ozo-config.cmake` should do the trick, but I'd need to have a closer look at this.

elsid:
I agree this is an issue. The problem is that ozo can't be build by Clang with `-Werror` and without `-Wno-gnu-string-literal-operator-template -Wno-gnu-zero-variadic-macro-arguments` in the current state. With GCC >= 9 it's also not possible but there are no similar to flags to Clang. So first we need to provide a way to build it by default without usage of string literal operator templates. One possible solution is to provide a marco like Boost.Hana [does](https://www.boost.org/doc/libs/1_75_0/libs/hana/doc/html/structboost_1_1hana_1_1string.html#a7f15faa77e126e1c2b08a2224ceefb84).

Regadrind the build system at Yandex. They don't use CMake, at least when I started to use ozo we had ya.make build system, you may find examples how it's used in [catboost](https://github.com/catboost/catboost/blob/master/catboost/app/ya.make) project. So basically everything CMake related is not for Yandex internal usage.

JonasProgrammer:
Ok, thank you for the insight. Adding warning options to the libraries interface is still a little debatable (after all, building with `-Werror` is up to the consumer), but I agree the current behavior should be retained and kept, as it is what's 'out there'.

[...]

PR #282, opened 2021-01-23
Conan support (rudimentary)

JonasProgrammer:
Hi,

sorry this took a little while, but here is finally a rudimentary conan integration.

The package aims to be compatible with both the `cmake_paths` generator, exporting the existing `ozo-config.cmake`, `ozo-targets.cmake` etc. files, as well as `cmake_find_package`, mirroring relevant definitions and dependencies from `CMakeLists.txt` in the conan components.

The reason for this double-specification is, that the central conan repository does not allow cmake config files to be exported, so this is my attempt at keeping it possible to be used in both scenarios. Potential further conan-center compatibility is also why I decided to pull in `resource_pool` as a package itself.

Please note, that this then still does not fully comply with the requirements for `conan-center`, I just tried to keep it in a way, so a migration further down the line is easy. Especially in regards to versioning and releasing versions, there is some more discussion needed, before we can try and get them included there.


JonasProgrammer:
Well sorry for this build spam, getting everything to run on travis was a little harder than anticipated

conanfile.py
```diff
@@ -0,0 +1,56 @@
...
+        version = re.search(r"^\s*project\(ozo\s+VERSION\s+([^\s)]+)", content, re.M).group(1)
+        return version.strip()
+    except Exception:
+        return None
+
+
+class Ozo(ConanFile):
+    name = 'ozo'
+    version = get_version()
+    license = 'MIT'
```
elsid: ozo has PostgreSQL License

elsid (changes requested):
A separate docker image for conan looks like an optimizaiton but this creates complications for the whole setup. I don't see how `docker_conan clang release` will work and `docker clang conan` is also incorrect. But still build.sh help suggest that these commands are correct. Can't we install all conan related packages into the existing build image? Generally the idea for this image is to have it built on your machine for ozo development, update it very rarely and don't remove it. With a separate conan image to test all possible changes locally you'd need two images.

conanfile.py
JonasProgrammer: My bad, that's what you get from copying and pasting the recipe instead of writing it from the ground up.

JonasProgrammer:
I made some further changes to incorporate fixes to #280, as the conan package should also add those options conditionally, depending on compiler and compiler version. Looking at [other examples](https://github.com/conan-io/conan-center-index/blob/689e2663fe41689259f61a5dbd3fe245e24c03b9/recipes/mp-units/all/conanfile.py#L64) from `conan-center`, this behaviour seems to be expected.

scripts/build.sh
```diff
@@ -195,9 +223,9 @@ case "$1" in
         build_${1}
         exit 0
     ;;
-    docker)
+    docker|docker_conan)
```
elsid: `docker_conan` is not used anymore.

scripts/build.sh
```diff
@@ -195,9 +223,9 @@ case "$1" in
...
         set -x
-        launch_in_docker $2 $3 $4 $5
+        launch_in_${1} $2 $3 $4 $5
```
elsid: This change is not required anymore.

elsid (approved):

thed636:
Guys, thanks for the contribution!

PR #284, opened 2021-02-10
Issue-279: Change is_bad predicate logic for connection

### LINKED ISSUE #279: Good / bad connection
I noticed that it seems to be a bit hard to interpret what it means for a bad vs a good connection.  This affects how we reuse connection and connection pool.

The codes below showed the interesting part.

The line 15 uses a fake port to connect to postgres.  The line 16 is an invalid database uri.

The codes intentionally didn't write commit() because anyway it will always fail.  It tries to start a transaction, and because the database is not there, it will certainly fail.  Then in the begin() callback, we rollback the transaction.  This rollback certainly will also fail (which is fine), but it will still have the rollback callback.  In that callback, I check the connection status.

It turns out that if we use line 15, the connection status is always Not Bad.  If we use line 16, it is indeed bad.

The same thing happens even if we don't use connection pool.  If we just open a simple connection, it is the same behavior.  This seems to tell that we cannot be based on connection status to decide if the connection is healthy (to reuse) or not.

```
   1 #include <boost/asio/io_service.hpp>
   2 #include <ozo/transaction.h>
   3 #include <ozo/connection_info.h>
   4 #include <ozo/connection_pool.h>
   5 #include <boost/asio.hpp>
   6 #include <boost/bind.hpp>
   7 #include <boost/date_time/posix_time/posix_time.hpp>
   8
   9 namespace asio = boost::asio;
  10
  11 using ConnectionPool = ozo::connection_pool<ozo::connection_info<>>;
  12 using Connection = ozo::connection_provider<ConnectionPool>::connection_type;
  13 using Transaction = ozo::transaction<Connection, decltype(ozo::make_options())>;
  14                                                                                                                                                                       no type named 'connection_info' in namespace 'ozo' [2]
  15 ozo::connection_info conn_info("postgresql://postgres:postgres@localhost:1000/test");
  16 // ozo::connection_info conn_info("abc");
  17 ozo::connection_pool_config conn_pool_config{10, 10, std::chrono::seconds(60), std::chrono::hours(24)};
  18 ConnectionPool conn_pool(conn_info, conn_pool_config);
  19 asio::io_context io;
  20
  21 void InTrans(const ozo::error_code& ec, Transaction tran) {
  22     if (ec) {
  23         std::cout << ec.message().c_str() << std::endl;
  24         std::cout << "Start to rollback" << std::endl;
  25         ozo::rollback(std::move(tran), [](ozo::error_code /*ec*/, Connection connection) {
  26             if (connection->is_bad()) {
  27                 std::cout << "is bad" << std::endl;
  28             } else {
  29                 std::cout << "is good" << std::endl;
  30             }
  31         });
  32     }
  33     std::cout << "in trans" << std::endl;
  34 }
  35
  36 void test() {
  37     using namespace std::chrono_literals;
  38     for (int i = 0; i < 2; ++i) {
  39         std::cout << "hello" << std::endl;
  40         ozo::begin(conn_pool[io], 3s, [](ozo::error_code ec, Transaction tran) {
  41             InTrans(ec, std::move(tran));
  42         });
  43     }
  44 }
  45
  46 int main() {
  47     test();
  48     io.run();
  49     return 0;
  50 }
```

Use line 15:
```
hello
hello
Connection timed out
Start to rollback
is good
in trans
Connection timed out
Start to rollback
is good
in trans
```

Use line 16:
```
hello
hello
pq_connection_status_bad - PQstatus returned CONNECTION_BAD
Start to rollback
is bad
in trans
pq_connection_status_bad - PQstatus returned CONNECTION_BAD
Start to rollback
is bad
in trans
```

Timur-Ziganshin:


include/ozo/query_builder.h
```diff
@@ -183,6 +183,7 @@ struct get_query_params_impl<query_builder<Ts...>> {
 
 namespace literals {
 
+# pragma GCC diagnostic ignored "-Wpedantic"
```
thed636: I'm not sure that this is a good choice to ignore the `pedantic` here without any push/pop. All the tests are green for the master. What problem does this solve?

include/ozo/impl/connection.h
```diff
@@ -85,7 +85,7 @@ void connection<OidMap, Statistics>::cancel() noexcept {
 
 template <typename OidMap, typename Statistics>
 bool connection<OidMap, Statistics>::is_bad() const noexcept {
-    return detail::connection_status_bad(native_handle());
+    return !native_handle() || PQstatus(native_handle()) != CONNECTION_OK;
```
thed636: IMO it is better to make a complementary function `detail::connection_status_ok()` in the same way that `detail::connection_status_bad()` implemented and use it negation in both places to preserve the logic and semantic if you do not mind.

thed636 (changes requested):
Hi Timur!

Thanks for the contribution! In addition to the requested changes, please be kind to add yourself to the [AUTHORS](https://github.com/yandex/ozo/blob/master/AUTHORS) file as the other contributors did.

include/ozo/query_builder.h
Timur-Ziganshin: I have problems with compilation without this line. Agreed with push/pop. I will try another time in docker

include/ozo/impl/connection.h
Timur-Ziganshin: Ok, but if we are going to implement `detail::connection_status_ok()` , it should be like this: `native_handle() && PQstatus(native_handle()) == CONNECTION_OK` , and its negation would not be what we guess

thed636: > its negation would not be what we guess

Well, as for me, everything is ok:
`native_handle() && PQstatus(native_handle()) == CONNECTION_OK`

The negation is:
`!(native_handle() && PQstatus(native_handle()) == CONNECTION_OK)`

equals to:
`!native_handle() || !(PQstatus(native_handle()) == CONNECTION_OK)`

equals to the proposed logic:
`!native_handle() || PQstatus(native_handle()) != CONNECTION_OK`

Am I wrong here?

include/ozo/query_builder.h
thed636: If you have a problem building without docker, could you share the compiler version which causes the problem? In a normal case, there should not be any difference between the docker image building and the host OS building with the same compiler flags.

Timur-Ziganshin: gcc 9.3.0, ok with clang.
I will remove this

include/ozo/impl/connection.h
Timur-Ziganshin: No, now i see, thx

thed636 (approved):

PR #289, opened 2021-02-28
Add more usage examples

elsid:
To demonstrate on practice more library features:
* Request with callback completion token
* Request with future completion token
* Retries with connection pool with simulated network disruption 
* Composite types usage


thed636:
👍

PR #296, opened 2021-06-04
Remove deprecated std::allocator<void>

thed636:
Clang claims about std::allocator<void> is deprecated in C++17

```
Apple clang version 12.0.5 (clang-1205.0.22.9)
Target: arm64-apple-darwin20.5.0
```

### DIFF
--- a/tests/test_asio.h
+++ b/tests/test_asio.h
@@ -119,20 +119,20 @@ struct execution_context : asio::execution_context {
 
         void on_work_finished() const {}
 
-        template <typename Function>
-        void dispatch(Function&& f, std::allocator<void>) const {
+        template <typename Function, typename Allocator>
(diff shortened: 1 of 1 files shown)


elsid:
Looks like boost is not longer available to download at https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.gz so all linux builds have failed. Need to switch to https://boostorg.jfrog.io/artifactory/main/release/1.66.0/source/boost_1_66_0.tar.gz . Or maybe setup our own mirror to avoid such issue in the future.

thed636:
Outdated versions are such a pain. :(

elsid:
It's not only outdated version. All boost.org downloads have been moved to boostorg.jfrog.io.

